PR #1, opened 2020-04-01
Add basic cmd and pkg for regitstration controller/agent

qiujian16:
This is to add leader election, healthz, metrics and event


cmd/registration/main.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	goflag "flag"
+	"fmt"
+	"math/rand"
+	"os"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
```
deads2k, line 12: @pmorie you think the gaps here still have value?  In the past go imports didn't handle them that well.

pmorie: I like the gaps, but i've found in large projects they're so inconsistent that they don't necessarily add a lot of value as a convention

pmorie:
This built for me locally, I'll merge and add the build machinery on top

PR #2, opened 2020-04-01
Add build machinery

pmorie:
Set up build machinery in order to set up prow


pmorie:
Green-buttoning so that the prow rehearse can run; I've tried the required make targets locally.

PR #7, opened 2020-04-08
implement the hub controller to reconcile SpokeCluster

skeeey:
After cluster-admin accept the spoke cluster, we will
1. create clusterrolebinding on the hub cluster.
2. create a namespace as the name of cluster on the hub cluster.
3. create role and rolebinding on the cluster namespace on the hub.


pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,367 @@ package spokecluster
...
+			if _, err := c.kubeClient.RbacV1().Roles(spokeCluster.Name).Create(ctx, &rbacv1.Role{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            spokeCluster.Name,
+					OwnerReferences: []metav1.OwnerReference{*metav1.NewControllerRef(spokeCluster, spokeClusterGVK)},
+				},
+				Rules: []rbacv1.PolicyRule{
+					{
+						Verbs:     []string{"create", "get", "list", "watch"},
+						APIGroups: []string{""},
+						Resources: []string{"namespaces"},
```
qiujian16: I do not think it should have this role. Maybe leave it empty, and add a TODO here?

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,367 @@ package spokecluster
...
+		klog.Infof("SpokeCluster %s is dennied", spokeCluster.Name)
+	}
+	return nil
+}
+
+func (c *Controller) reconcileAcceptedSpokeCluster(ctx context.Context, spokeCluster *v1.SpokeCluster) error {
+	// Current spoke cluster has not been accepted yet, we accecpt it
+	acceptedCondition := getSpokeClusterCondition(&spokeCluster.Status, v1.SpokeClusterConditionHubAccepted)
+	if acceptedCondition == nil {
+		if err := c.acceptSpokeCluster(ctx, spokeCluster); err != nil {
```
qiujian16: I think we can directly create namespace, role, rolebinding here instead of a wrapper func?

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,367 @@ package spokecluster
...
+	// Current spoke cluster has not been accepted yet, do nothing
+	acceptedCondition := getSpokeClusterCondition(&spokeCluster.Status, v1.SpokeClusterConditionHubAccepted)
+	if acceptedCondition == nil {
+		return nil
+	}
+
+	deniedCondition := getSpokeClusterCondition(&spokeCluster.Status, v1.SpokeClusterConditionHubDenied)
+	// Current spoke cluster is not denied, we deny it
+	if deniedCondition == nil {
+		if err := c.cleanupAcceptededSpokeCluster(ctx, spokeCluster.Name); err != nil {
```
qiujian16: i think you can directly delete role/rolebinding here instead of using a wrapper func

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,345 @@ package spokecluster
...
+			{
+				Kind:     "Group",
+				APIGroup: rbacv1.GroupName,
+				Name:     groupName,
+			},
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "ClusterRole",
+			Name:     clusterRoleName,
```
qiujian16: I think we also need to create a clusterrole to allow spoke agent update status, we may also need a rule in clusterrole to allow subresource under csr so the we can use that subresource to check whether agent can renew csr.

skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
 
-// Controller manages the SpokeCluster resources
-type Controller struct {
-	// TODO add client and other configurations
+const (
+	groupPrefix       = "system:open-cluster-management"
+	clusterRolePrefix = "system:open-cluster-management:spokecluster"
+)
+
+type spokeClusterController struct {
```
pmorie, line 31: +1 on the name change

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+		}
+		acceptedCondition.Status = metav1.ConditionFalse
+		acceptedCondition.LastTransitionTime = metav1.Now()
+		acceptedCondition.Reason = "HubClusterAdminDenied"
+		acceptedCondition.Message = "Denied by hub cluster admin"
+		_, err := c.clusterClient.ClusterV1().SpokeClusters().UpdateStatus(ctx, spokeCluster, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+		klog.Infof("SpokeCluster %s is denied", spokeCluster.Name)
```
pmorie: Consider emitting an event here.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+
+	// Current spoke cluster is accepted, do nothing.
+	if acceptedCondition != nil && acceptedCondition.Status == metav1.ConditionTrue {
+		return nil
+	}
+
+	// Hub cluster-admin accepts the spoke cluster, we create
+	// 1. clusterrole and clusterrolebinding for this spoke cluster.
+	// 2. namespace for this spoke cluster.
+	// 3. role and rolebindg for this spoke cluster on its namespace.
```
pmorie: typo: rolebinding

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+		_, err := c.clusterClient.ClusterV1().SpokeClusters().UpdateStatus(ctx, spokeCluster, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+		klog.Infof("SpokeCluster %s is denied", spokeCluster.Name)
+		return nil
+	}
+
+	// Current spoke cluster is accepted, do nothing.
+	if acceptedCondition != nil && acceptedCondition.Status == metav1.ConditionTrue {
```
pmorie: What happens if the clusterrole or another resource related to the spoke within the hub is accidentally deleted out of band?

pmorie: I would expect that if one of the resources was somehow deleted out of band, that they would be re-applied if the spoke was still accepted, so that the spoke controller could recover.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+func (c *spokeClusterController) applySpokeClusterNamespace(ctx context.Context, spokeClusterName string) error {
+	_, err := c.kubeClient.CoreV1().Namespaces().Get(ctx, spokeClusterName, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			if _, err := c.kubeClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{Name: spokeClusterName},
+			}, metav1.CreateOptions{}); err != nil {
+				return err
+			}
+			klog.Infof("The spoke cluster namespace %s is created.", spokeClusterName)
```
pmorie: This would be a useful event to have

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+		return err
+	}
+	klog.Infof("The role %s is deleted from namespace %s.", spokeClusterName, spokeClusterName)
+	return nil
+}
+
+func (c *spokeClusterController) applyRoleBinding(ctx context.Context, spokeClusterName string) error {
+	roleBinding := &rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{Name: spokeClusterName},
+		Subjects: []rbacv1.Subject{
```
pmorie: I could see group access as valid here, but it deserves a comment in the code to explain why group or user is chosen.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+				Resources: []string{"certificatesigningrequests/spokeclusteragent"},
+				Verbs:     []string{"create"},
+			},
+			// Allow spoke agent to update the status of its owner spoke cluster
+			{
+				APIGroups:     []string{"cluster.open-cluster-management.io"},
+				Resources:     []string{"spokeclusters/status"},
+				ResourceNames: []string{spokeClusterName},
+				Verbs:         []string{"patch", "update"},
+			},
```
pmorie: Likely missing a couple of rules here. The spoke agent needs to be able to list/watch/get.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+		return err
+	}
+	klog.Infof("The spoke cluster namespace %s is deleted.", spokeClusterName)
+	return nil
+}
+
+func (c *spokeClusterController) applyRole(ctx context.Context, spokeClusterName string) error {
+	role := &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{Name: spokeClusterName},
+		//TODO: add rules later
```
pmorie: I think this is all that you can do for now (leaving a todo)

pkg/hub/spokecluster/controller_test.go
```diff
@@ -0,0 +1,201 @@
...
+		return err
+	}
+
+	if len(spokeCluster.Status.Conditions) != 1 {
+		return fmt.Errorf("expacted one condition, but got %d", len(spokeCluster.Status.Conditions))
+	}
+
+	acceptedCondition := getSpokeClusterCondition(&spokeCluster.Status, v1.SpokeClusterConditionHubAccepted)
+	if acceptedCondition == nil {
+		return fmt.Errorf("expact there is an accepted condition, but failed")
```
pmorie: typo: expect

pkg/hub/spokecluster/controller_test.go
```diff
@@ -0,0 +1,201 @@
...
+				if err := verifySpokeClusterNamespace(false); err != nil {
+					t.Errorf("unexpected err: %v", err)
+				}
+				if err := verifySpokeClusterCondition(metav1.ConditionTrue); err != nil {
+					t.Errorf("unexpected err: %v", err)
+				}
+			},
+		},
+		{
+			"deniy the SpokeCluster again",
```
pmorie: typo: deny

pkg/hub/spokecluster/controller_test.go
```diff
@@ -0,0 +1,201 @@
...
+func TestSyncSpokeCluster(t *testing.T) {
+	ctrl := spokeClusterController{fakeKubeClient, fakeClusterClient}
+
+	cases := []struct {
+		desc   string
+		before func(t *testing.T)
+		after  func(t *testing.T)
+	}{
+		{
+			"sync a deleted spoke cluster",
```
pmorie: this test case looks incomplete

pkg/hub/spokecluster/controller_test.go
```diff
@@ -0,0 +1,201 @@
...
+			func(t *testing.T) {},
+			func(t *testing.T) {},
+		},
+		{
+			"create a new spoke cluster",
```
pmorie: tip: use names of fields here, it will make this much easier to read accurately

pkg/hub/spokecluster/controller_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	spokeCluster.Spec.HubAcceptsClient = hubAcceptsClient
+	_, err = fakeClusterClient.ClusterV1().SpokeClusters().Update(context.TODO(), spokeCluster, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func verifySpokeClusterNamespace(expectedNotFound bool) error {
```
pmorie: I would change the semantics of this method slightly to be

`func verifySpokeClusterNamespace(expected bool) error {`

so that the value passed is align with whether the resource is expected to exist, not its opposite

pmorie (commented):
Initial pass, looks pretty good but left a couple comments

pkg/hub/spokecluster/controller.go
skeeey: yes, I  will remove this judgment at here, the subsequent actions are apply methods, so if the resources is accidentally deleted, they will be recovered

pkg/hub/spokecluster/controller.go
skeeey: At this time point, I think the spoke agent already be able to list/watch/get.

pkg/hub/spokecluster/controller.go
skeeey: ok, I will emit an event here

pkg/hub/spokecluster/controller.go
skeeey: ok, I will emit an event here too

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller_test.go
skeeey: done

pkg/hub/spokecluster/controller_test.go
skeeey: done

pkg/hub/spokecluster/controller_test.go
skeeey: done

pkg/hub/spokecluster/controller_test.go
skeeey: done

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
 		WithSync(c.sync).
-		ResyncEvery(time.Minute).
 		ToController("SpokeClusterController", recorder)
 }
 
-func (c Controller) sync(ctx context.Context, syncCtx factory.SyncContext) error {
-	// TODO add main reconcile loop
+func (c *spokeClusterController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	spokeClusterName := syncCtx.QueueKey()
+	klog.Infof("Reconciling SpokeCluster %s", spokeClusterName)
```
deads2k: V(4)

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+	spokeCluster, err := c.clusterClient.ClusterV1().SpokeClusters().Get(ctx, spokeClusterName, metav1.GetOptions{})
+	if err != nil {
```
deads2k:
```go
if notFound(){
}
if err != nil{
}
```

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+		if errors.IsNotFound(err) {
+			// Spoke cluster not found, could have been deleted, we remove its related resources.
+			return c.removeSpokeClusterResources(ctx, spokeClusterName)
```
deads2k: This indicates a need to add a finalizer on the resource when you accept them.  @pmorie can we get an enhancements repo open?

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+	klog.Infof("The clusterrolebinding %s is deleted.", clusterRoleBindingName)
+	return nil
+}
+
+func getSpokeClusterCondition(status *v1.SpokeClusterStatus, conditionType string) *v1.StatusCondition {
```
deads2k: create helpers that look like: https://github.com/openshift/library-go/blob/master/pkg/operator/v1helpers/helpers.go#L49-L108

we will be moving functions like that into kube/apimachinery eventually.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+			// Spoke cluster not found, could have been deleted, we remove its related resources.
+			return c.removeSpokeClusterResources(ctx, spokeClusterName)
+		}
+		return err
+	}
+
+	acceptedCondition := getSpokeClusterCondition(&spokeCluster.Status, v1.SpokeClusterConditionHubAccepted)
+	if !spokeCluster.Spec.HubAcceptsClient {
+		// Current spoke cluster is not accepted, do nothing.
+		if acceptedCondition == nil || acceptedCondition.Status == metav1.ConditionFalse {
```
deads2k: this block looks incorrect.  the spoke isn't accepted, run through the removed logic again.  If this proves too expensive, we'll make a condition for "controller has removed spoke"

deads2k: > this block looks incorrect. the spoke isn't accepted, run through the removed logic again. If this proves too expensive, we'll make a condition for "controller has removed spoke"

hmmm... I take that back. I think this logic is ok-ish.   it would take someone manually fiddling with this *after* you create the finalizer.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+
+		// Hub cluster-admin denies the current spoke cluster, we remove its related resources and update its condition.
+		if err := c.removeSpokeClusterResources(ctx, spokeCluster.Name); err != nil {
+			return err
+		}
+		acceptedCondition.Status = metav1.ConditionFalse
+		acceptedCondition.LastTransitionTime = metav1.Now()
+		acceptedCondition.Reason = "HubClusterAdminDenied"
+		acceptedCondition.Message = "Denied by hub cluster admin"
+		_, err := c.clusterClient.ClusterV1().SpokeClusters().UpdateStatus(ctx, spokeCluster, metav1.UpdateOptions{})
```
deads2k: we've been very happy with logic like this: https://github.com/openshift/library-go/blob/master/pkg/operator/v1helpers/helpers.go#L150-L190

pkg/hub/spokecluster/controller.go
deads2k: > I would expect that if one of the resources was somehow deleted out of band, that they would be re-applied if the spoke was still accepted, so that the spoke controller could recover.

+1

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+	// Current spoke cluster is accepted, do nothing.
+	if acceptedCondition != nil && acceptedCondition.Status == metav1.ConditionTrue {
+		return nil
+	}
+
+	// Hub cluster-admin accepts the spoke cluster, we create
+	// 1. clusterrole and clusterrolebinding for this spoke cluster.
+	// 2. namespace for this spoke cluster.
+	// 3. role and rolebindg for this spoke cluster on its namespace.
+	if err := c.applyClusterRole(ctx, spokeCluster.Name); err != nil {
```
deads2k: consider using a library like https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceread/rbac.go#L20-L26 and https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/rbac.go#L20 .

deads2k: using manifests means that you can easily test manually and we can edit them in a way that is "normal"

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+		acceptedCondition.Status = metav1.ConditionTrue
+		acceptedCondition.LastTransitionTime = metav1.Now()
+		acceptedCondition.Reason = "HubClusterAdminAccepted"
+		acceptedCondition.Message = "Accepted by hub cluster admin"
+	}
+	_, err = c.clusterClient.ClusterV1().SpokeClusters().UpdateStatus(ctx, spokeCluster, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	klog.Infof("SpokeCluster %s is accepted", spokeCluster.Name)
```
deads2k: event

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,411 @@ package spokecluster
...
+	return nil
+}
+
+func (c *spokeClusterController) removeSpokeClusterResources(ctx context.Context, spokeClusterName string) error {
+	if err := c.removeRoleBinding(ctx, spokeClusterName); err != nil {
```
deads2k: namespaced resources don't need to be removed individually.

pkg/hub/spokecluster/controller_test.go
```diff
@@ -0,0 +1,201 @@
...
+	fakeClusterClient = clusterfake.NewSimpleClientset()
+	fakeKubeClient    = kubefake.NewSimpleClientset()
+)
+
+func TestSyncSpokeCluster(t *testing.T) {
+	ctrl := spokeClusterController{fakeKubeClient, fakeClusterClient}
+
+	cases := []struct {
+		desc   string
+		before func(t *testing.T)
```
deads2k: we normally  have a function to produce the initial objects (not an anything func) `initialObjects() []runtime.Object`.  Then we have a `validActions(t, fakeClient)` function and an expectedErr function. That gives a lot more control/confidence in the tests.

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: cool, that is a good example

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: cool, if there is a library, it is best

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: > can we get an enhancements repo open?

@deads2k do you mean we will have a library to support  finalizer? I implement the finalizer firstly

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller_test.go
skeeey: +1, refactor the test

pkg/hub/spokecluster/manifests/spokecluster-clusterrole.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:open-cluster-management:spokecluster:{{ .SpokeClusterName }}
+rules:
+# Allow spoke agent to renew its certificate
+- apiGroups: ["certificates.k8s.io"]
+  resources: ["certificatesigningrequests/spokeclusteragent"]
```
deads2k, line 8: I may need to go back through, but shouldn't this simply be get, list, watch, create on certificatesigningrequests?

pkg/hub/spokecluster/manifests/spokecluster-clusterrole.yaml
```diff
@@ -0,0 +1,13 @@
...
+  verbs: ["create"]
+# Allow spoke agent to update the status of its owner spoke cluster
```
deads2k, line 15: I expected to see this restricted by name

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
+	"k8s.io/klog"
 )
 
-// Controller manages the SpokeCluster resources
-type Controller struct {
-	// TODO add client and other configurations
+const (
+	manifestDir           = "pkg/hub/spokecluster"
+	clusterRolePrefix     = "system:open-cluster-management:spokecluster"
+	spokeClusterFinalizer = "cluster.open-cluster-management.io/deletion-protection"
```
deads2k: cluster.open-cluster-management.io/api-resource-cleanup

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
+}
+
+func (c *spokeClusterController) removeSpokeClusterFinalizer(ctx context.Context, spokeClusterName string) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		spokeCluster, err := c.clusterClient.ClusterV1().SpokeClusters().Get(ctx, spokeClusterName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		finalizers := sets.NewString(spokeCluster.Finalizers...)
```
deads2k: technically this is ok, but some finalizers are dumb and want stable ordering.  Please build up an array that keeps the existing order.  This is general practice for finalizers.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
+
+	clusterRoleName := fmt.Sprintf("%s:%s", clusterRolePrefix, spokeClusterName)
+	err = c.kubeClient.RbacV1().ClusterRoles().Delete(ctx, clusterRoleName, metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	c.eventRecorder.Eventf("SpokeClusterClusterRoleDeleted", "clusterrole %s is deleted", clusterRoleName)
+
+	clusterRoleBindingName := fmt.Sprintf("%s:%s", clusterRolePrefix, spokeClusterName)
+	err = c.kubeClient.RbacV1().ClusterRoleBindings().Delete(ctx, clusterRoleBindingName, metav1.DeleteOptions{})
```
deads2k, line 193: This won't age well.  Add a TODO to search all clusterroles and roles for this group and remove the entry or delete the clusterrolebinding if its the only subject.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
+	klog.V(4).Infof("Reconciling SpokeCluster %s", spokeClusterName)
+	spokeCluster, err := c.clusterClient.ClusterV1().SpokeClusters().Get(ctx, spokeClusterName, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		// Spoke cluster not found, could have been deleted, do nothing.
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
```
deads2k, line 68: start this function by adding the finalizer if required and return.  We will be retriggered when the update is observed.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
 }
+
+func (c *spokeClusterController) addSpokeClusterFinalizer(ctx context.Context, spokeClusterName string) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		spokeCluster, err := c.clusterClient.ClusterV1().SpokeClusters().Get(ctx, spokeClusterName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		finalizers := sets.NewString(spokeCluster.Finalizers...)
```
deads2k: just add yourself to the end if needed.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	c.eventRecorder.Eventf("SpokeClusterClusterRoleBindingDeleted", "clusterrolebinding %s is deleted", clusterRoleBindingName)
+
 	return nil
 }
+
+func (c *spokeClusterController) addSpokeClusterFinalizer(ctx context.Context, spokeClusterName string) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
deads2k: don't retry on conflict for this one. KISS

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
+		}
+		spokeCluster.Finalizers = append(spokeCluster.Finalizers, spokeClusterFinalizer)
+
+		_, err = c.clusterClient.ClusterV1().SpokeClusters().Update(ctx, spokeCluster, metav1.UpdateOptions{})
+		return err
+	})
+}
+
+func (c *spokeClusterController) removeSpokeClusterFinalizer(ctx context.Context, spokeClusterName string) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
deads2k: don't retry on conflict for this one. KISS  We'll simply try again later.

deads2k: don't retry on conflict for this one. KISS  We'll simply try again later.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
+		}
+
+		return nil
+	}
+
+	if !spokeCluster.Spec.HubAcceptsClient {
+		acceptedCondition := helpers.FindSpokeClusterCondition(spokeCluster.Status.Conditions, v1.SpokeClusterConditionHubAccepted)
+		// Current spoke cluster is not accepted, do nothing.
+		if acceptedCondition == nil || acceptedCondition.Status == metav1.ConditionFalse {
+			return err
```
deads2k: this isn't right.  err has something unrelated here, right?

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
+		}
+
+		return nil
+	}
+
+	if !spokeCluster.Spec.HubAcceptsClient {
+		acceptedCondition := helpers.FindSpokeClusterCondition(spokeCluster.Status.Conditions, v1.SpokeClusterConditionHubAccepted)
+		// Current spoke cluster is not accepted, do nothing.
+		if acceptedCondition == nil || acceptedCondition.Status == metav1.ConditionFalse {
```
deads2k: seems like this is  `!IsConditionTrue` (note that it is not the same as isConditionFalse).  There should be a helper for that

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
+				Type:    v1.SpokeClusterConditionHubAccepted,
+				Status:  metav1.ConditionFalse,
+				Reason:  "HubClusterAdminDenied",
+				Message: "Denied by hub cluster admin",
+			}),
+		)
+		if err != nil {
+			return err
+		}
+		c.eventRecorder.Eventf("SpokeClusterDenied", "spoke cluster %s is denied by hub cluster admin", spokeClusterName)
```
deads2k: move the event to before you start removing stuff.  Just in case of failures.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,231 @@ package spokecluster
...
+		"manifests/spokecluster-rolebinding.yaml",
+	)
+	errs := []error{}
+	for _, result := range resourceResults {
+		if result.Error != nil {
+			errs = append(errs, fmt.Errorf("%q (%T): %v", result.File, result.Type, result.Error))
+		}
+	}
+	if err := operatorhelpers.NewMultiLineAggregate(errs); err != nil {
+		return err
```
deads2k: we need to indicate this error on the `SpokeClusterStatus`

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: yes, this isn't right, corrected

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: added

pkg/hub/spokecluster/manifests/spokecluster-clusterrole.yaml
skeeey: added the resource name in the rule

pkg/hub/spokecluster/controller.go
skeeey: changed

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: changed

pkg/hub/spokecluster/controller.go
skeeey: changed

pkg/hub/spokecluster/controller.go
skeeey: added

pkg/hub/spokecluster/manifests/spokecluster-clusterrole.yaml
skeeey: This rule is used to allow spoke agent rotate its cert after its cert is out of date, I plan to use the SubjectAccessReview API to check whether a spoke agent can rotate its cert when hub controller need auto approve a rotation cert

And we will create a clusterrole for a spoke agent after its csr  is approved by hub cluster admin, the clusterrole will have two rules:
- get, list, watch, create on certificatesigningrequests
- get, list, watch, create on spokeclusters 

This will be implemented in csr controller

so at this time point, the spoke agent already has the permission to get, list, watch, create on certificatesigningrequests

skeeey: Or, I keep the get, list, watch, create on certificatesigningrequests and spokeclusters and make sure this role name is same as the csr controller created, this means

- if the csr controller created role existed, I update it
- if the csr controller created role was deleted accidentally, I recover it

and I also need to change the the remove clusterrole/clusterrolebinding action to an update action

I think this may be more reasonable, what's your suggestion? @deads2k

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,222 @@ package spokecluster
...
+	if errors.IsNotFound(err) {
+		// Spoke cluster not found, could have been deleted, do nothing.
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	// Hub cluster-admin accepts the spoke cluster, we add the finalizer if it does not exist firstly
+	if spokeCluster.Spec.HubAcceptsClient && spokeCluster.DeletionTimestamp.IsZero() {
```
deads2k: no need to check `spokeCluster.Spec.HubAcceptsClient`.  Just check to be sure it isn't deleted.  We want this applied early.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,222 @@ package spokecluster
...
+		}
+
+		// Hub cluster-admin denies the current spoke cluster, we remove its related resources and update its condition.
+		c.eventRecorder.Eventf("SpokeClusterDenied", "spoke cluster %s is denied by hub cluster admin", spokeClusterName)
+
+		if err := c.removeSpokeClusterResources(ctx, spokeClusterName); err != nil {
+			return err
+		}
+
+		if err := c.removeSpokeClusterFinalizer(ctx, spokeCluster); err != nil {
```
deads2k: don't remove the finalizer. We'll want another shot at it when it is deleted.  Just to be sure.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,222 @@ package spokecluster
...
+		if err := c.removeSpokeClusterResources(ctx, spokeClusterName); err != nil {
+			return err
+		}
+		return c.removeSpokeClusterFinalizer(ctx, spokeCluster)
+	}
+
+	if !spokeCluster.Spec.HubAcceptsClient {
+		acceptedCondition := helpers.FindSpokeClusterCondition(spokeCluster.Status.Conditions, v1.SpokeClusterConditionHubAccepted)
+		// Current spoke cluster is not accepted, do nothing.
+		if !helpers.IsConditionTrue(acceptedCondition) {
```
deads2k, line 95: sorry about this.  Reading through I think we actually want this condition to be false.  if it isn't false, we want to do the delete procedure and set it.

pkg/hub/spokecluster/controller.go
```diff
@@ -2,27 +2,222 @@ package spokecluster
...
+	acceptedCondition := v1.StatusCondition{
+		Type:    v1.SpokeClusterConditionHubAccepted,
+		Status:  metav1.ConditionTrue,
+		Reason:  "HubClusterAdminAccepted",
+		Message: "Accepted by hub cluster admin",
+	}
+
+	if len(errs) > 0 {
+		applyErrors := operatorhelpers.NewMultiLineAggregate(errs)
+		acceptedCondition.Reason = "Error"
```
deads2k, line 158: @pmorie i think this points to a need to have a degraded style condition here.  Do you agree?

pkg/hub/spokecluster/manifests/spokecluster-clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
...
+rules:
+# Allow spoke agent to rotate its certificate
+- apiGroups: ["certificates.k8s.io"]
+  resources: ["certificatesigningrequests/spokeclusteragent"]
+  verbs: ["create"]
+# Allow spoke agent to update the status of its owner spoke cluster
+- apiGroups: ["cluster.open-cluster-management.io"]
+  resources: ["spokeclusters/status"]
+  resourceNames: ["{{ .SpokeClusterName }}"]
+  verbs: ["patch", "update"]
```
deads2k, line 19: it needs get list watch on the particularly named spoke cluster (not status) too.

pkg/hub/spokecluster/manifests/spokecluster-clusterrole.yaml
deads2k: I'm just saying this permission doesn't do anything.

deads2k:
I'd like to see some followups, but we can start here.

/lgtm

pkg/hub/spokecluster/controller.go
pmorie: +1, agree

pkg/hub/spokecluster/controller.go
pmorie: Agree.

pkg/hub/spokecluster/controller.go
pmorie: Agree

pkg/hub/spokecluster/controller_test.go
```diff
@@ -0,0 +1,234 @@
...
+		expectedErr     string
+	}{
+		{
+			name:            "sync a deleted spoke cluster",
+			startingObjects: []runtime.Object{},
+			validateActions: func(t *testing.T, actions []clienttesting.Action) {
+				if len(actions) != 1 {
+					t.Errorf("expected 1 call but got: %#v", actions)
+				}
+				_, ok := actions[len(actions)-1].(clienttesting.GetActionImpl)
```
pmorie: I like the use of the fake client here. I would like to see additional detail about each action. Make a method like `assertAction(t, actions[0], clienttesting.GetActionImpl, reflect.TypeOf(v1.SpokeCluster{}))` so you can check some basics about each action.

pkg/hub/spokecluster/controller_test.go
```diff
@@ -0,0 +1,234 @@
...
+			validateActions: func(t *testing.T, actions []clienttesting.Action) {
+				if len(actions) != 2 {
+					t.Errorf("expected 2 call but got: %#v", actions)
+				}
+				lastAction, ok := actions[len(actions)-1].(clienttesting.UpdateActionImpl)
+				if !ok {
+					t.Errorf("expected get call but got: %#v", actions)
+				}
+				actual := lastAction.Object.(*v1.SpokeCluster)
+				if len(actual.Finalizers) != 1 {
```
pmorie: It might be worth a `assertFinalizer(t, actual, spokeClusterFinalizer)` method

pkg/hub/spokecluster/controller_test.go
```diff
@@ -0,0 +1,234 @@
...
+				}
+				lastAction, ok := actions[len(actions)-1].(clienttesting.UpdateActionImpl)
+				if !ok {
+					t.Errorf("expected update call but got: %#v", actions)
+				}
+				actual := lastAction.Object.(*v1.SpokeCluster)
+				if len(actual.Status.Conditions) != 1 {
+					t.Errorf("expected 1 condition but got: %#v", actual.Status.Conditions)
+				}
+				if actual.Status.Conditions[0].Status != metav1.ConditionTrue {
```
pmorie: You must test the condition type as part of this assertion and you may want to check the reason as well.

might be worth a `assertCondition(t, actual, v1.SpokeClusterConditionHubAccepted, ConditionTrue)` method

pkg/hub/spokecluster/controller_test.go
```diff
@@ -0,0 +1,234 @@
...
+				}
+				if actual.Finalizers[0] != spokeClusterFinalizer {
+					t.Errorf("expected %q but got %s", spokeClusterFinalizer, actual.Finalizers[0])
+				}
+			},
+		},
+		{
+			name:            "accept a spoke cluster",
+			startingObjects: []runtime.Object{newAcceptedSpokeCluster([]string{spokeClusterFinalizer})},
+			validateActions: func(t *testing.T, actions []clienttesting.Action) {
```
pmorie, line 54: sketch of a version of this method incorporating my other suggestions:

```go
func(t *testing.T, actions []clienttesting.Action) {
  if len(actions) != 3 {
    t.Errorf("...")
  }
  assertAction(t, actions[0], clienttesting.GetActionImpl, spokeClusterType)
  ...
  assertAction(t, actions[2], clienttesting.UpdateActionImpl, spokeClusterType)
  actual := actions[2].(clienttesting.UpdateActionImpl).Object
  assertCondition(t, actual, v1.SpokeClusterConditionHubAccepted, ConditionTrue)
}
```

pkg/hub/spokecluster/controller.go
skeeey: yes

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller.go
skeeey: done

pkg/hub/spokecluster/controller_test.go
skeeey: +1, this is more clear for test

pkg/hub/spokecluster/controller_test.go
skeeey: done

pkg/hub/spokecluster/controller_test.go
skeeey: go cannot use `type` as a param, so another `assertAction` method was given

pkg/hub/spokecluster/controller_test.go
skeeey: done

pkg/hub/spokecluster/manifests/spokecluster-clusterrole.yaml
skeeey: added

PR #6, opened 2020-04-03
Add agent bootstrap & cert-rotation

elgnay:
The unit test cases are not committed yet. I'll add them soon.

PR #10, opened 2020-04-20
Address the remaining comments on PR #6

elgnay:
This PR addressed the remaining comments on PR #6

### DIFF
--- a/pkg/spoke/hubclientcert/certificate.go
+++ b/pkg/spoke/hubclientcert/certificate.go
@@ -86,7 +86,7 @@ func getCertValidityPeriod(secret *corev1.Secret) (*time.Time, *time.Time, error
 
 	certs, err := certutil.ParseCertsPEM(certData)
 	if err != nil {
-		return nil, nil, fmt.Errorf("unable to parse TLS certificates: %v", err)
+		return nil, nil, fmt.Errorf("unable to parse TLS certificates: %w", err)
 	}
 
 	if len(certs) == 0 {
@@ -142,6 +142,7 @@ func buildKubeconfig(clientConfig *restclient.Config, certPath, keyPath string) 
 }
 
 func isCSRApproved(csr *certificates.CertificateSigningRequest) bool {
+	// TODO: need to make it work in csr v1 as well
 	approved := false
 	for _, condition := range csr.Status.Conditions {
 		if condition.Type == certificates.CertificateDenied {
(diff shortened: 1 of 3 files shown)


pkg/spoke/hubclientcert/certificate.go
```diff
@@ -86,7 +86,7 @@ func getCertValidityPeriod(secret *corev1.Secret) (*time.Time, *time.Time, error
 
 	certs, err := certutil.ParseCertsPEM(certData)
 	if err != nil {
-		return nil, nil, fmt.Errorf("unable to parse TLS certificates: %v", err)
+		return nil, nil, fmt.Errorf("unable to parse TLS certificates: %w", err)
```
pmorie, line 89: Why change this? `%w` will just be a synonym for `%v` here

pkg/spoke/spokeagent.go
```diff
@@ -69,15 +70,19 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
...
 	// create kube client for spoke cluster
 	spokeKubeClient, err := kubernetes.NewForConfig(controllerContext.KubeConfig)
 	if err != nil {
 		return err
 	}
 
+	// create an informer for secrets on spoke cluster
+	spokeSecretInformer := informers.NewSharedInformerFactory(spokeKubeClient,
+		10*time.Minute).Core().V1().Secrets()
+	go spokeSecretInformer.Informer().Run(ctx.Done())
```
pmorie, line 84: nit: I think the `Run` invocation could move next to where the other informer and controller are run

pmorie:
Just a couple nits, lgtm otherwise

pkg/spoke/hubclientcert/certificate.go
deads2k: > Why change this? `%w` will just be a synonym for `%v` here

it allows determining chained types, right?

deads2k: > Why change this? `%w` will just be a synonym for `%v` here

see the wrappping errors bits of https://blog.golang.org/go1.13-errors

deads2k:
/lgtm
/hold

holding for @pmorie's request

pkg/spoke/hubclientcert/certificate.go
pmorie: Interesting, I was not aware of this, disregard my comments re: `%w`

PR #13, opened 2020-04-27
enable open-cluster-management registration integration test

skeeey:


Makefile
```diff
@@ -17,3 +17,11 @@ clean:
 .PHONY: clean
 
 GO_TEST_PACKAGES :=./pkg/... ./cmd/...
+
+test-install-ginkgo:
+	go get github.com/onsi/ginkgo/ginkgo
```
pmorie, line 22: You don't need to pull the ginkgo binary to run the ginkgo tests, instead build the tests with `-c` and you will get a test binary that you can run the exact same way.

build/build-image/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM openshift/origin-release:golang-1.13
```
pmorie, line 1: It's not clear yet what this Dockerfile is for - why do we need it?

Makefile
skeeey: cool, using `go test -c`

build/build-image/Dockerfile
skeeey: I try to prepare the integration test env (download kubebuilder for kube-apiserver and etcd) with this Dockerfile, I can use make to replace it, thus we don't need this Dockerfile

PR #11, opened 2020-04-23
spoke agent spoke cluster controller

skeeey:

PR #8, opened 2020-04-13
hub csr controller

skeeey:
hub csr controller will
- create the clusterrole/clusterrolebinding after csr is approved
- auto approve the accepted spoke cluster cert rotation

PR #15, opened 2020-04-28
Resolve comments on PR #11 and #8

skeeey:


pkg/spoke/spokeagent.go
```diff
@@ -236,7 +236,7 @@ func (o *SpokeAgentOptions) AddFlags(fs *pflag.FlagSet) {
 		"The name of secret in component namespace storing kubeconfig for hub.")
 	fs.StringVar(&o.HubKubeconfigDir, "hub-kubeconfig-dir", o.HubKubeconfigDir,
 		"The mount path of hub-kubeconfig-secret in the container.")
-	fs.StringVar(&o.SpokeServerUrl, "spoke-server-url", o.SpokeServerUrl,
+	fs.StringVar(&o.SpokeServerUrl, "spoke-external-server-url", o.SpokeServerUrl,
```
pmorie: Let's change the name of to `o.SpokeExternalServerUrl` too

skeeey: ok, done

pkg/spoke/spokeagent.go
```diff
@@ -110,39 +108,69 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
...
+		caBundle := controllerContext.KubeConfig.CAData
+		if caBundle == nil && controllerContext.KubeConfig.CAFile != "" {
+			data, err := ioutil.ReadFile(controllerContext.KubeConfig.CAFile)
+			if err != nil {
+				return fmt.Errorf("unable to load CA data from file %q: %w", controllerContext.KubeConfig.CAFile, err)
+			}
+			caBundle = data
 		}
+
+		// TODO there is a corner case, if the hub kubeconfig is ready, but the spoke cluster did not create, and the agent is
```
deads2k, line 146: this is evidence that you really want this controller always running and just silent until it needs to do work.

pkg/spoke/spokecluster/creating_controller.go
```diff
@@ -0,0 +1,77 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// spokeClusterCreatingController creates a spoke cluster on hub cluster during the spoke agent bootstrap phase
+type spokeClusterCreatingController struct {
+	isSpokeClusterExists   bool
+	clusterName            string
+	spokeExternalServerUrl string
+	spokeCABundle          []byte
+	hasSpokeClusterFunc    func(bool)
```
deads2k, line 22: this looks suspicious

pkg/spoke/spokecluster/creating_controller.go
```diff
@@ -0,0 +1,77 @@
...
+		hubClusterClient:       hubClusterClient,
+	}
+	return factory.New().
+		WithSync(c.sync).
+		ResyncEvery(5*time.Minute).
+		ToController("SpokeClusterCreatingController", recorder)
+}
+
+func (c *spokeClusterCreatingController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	if c.isSpokeClusterExists {
```
deads2k, line 47: Do a lister get here

pkg/spoke/spokecluster/creating_controller.go
```diff
@@ -0,0 +1,77 @@
...
+	if errors.IsAlreadyExists(err) {
+		c.isSpokeClusterExists = true
+		c.hasSpokeClusterFunc(c.isSpokeClusterExists)
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("unable to create spoke cluster with name %q on hub: %w", c.clusterName, err)
+	}
+
+	c.isSpokeClusterExists = true
```
deads2k, line 73: you don't need this

pkg/spoke/spokecluster/creating_controller.go
```diff
@@ -0,0 +1,77 @@
...
+		hasSpokeClusterFunc:    hasSpokeClusterFunc,
+		hubClusterClient:       hubClusterClient,
+	}
+	return factory.New().
+		WithSync(c.sync).
+		ResyncEvery(5*time.Minute).
+		ToController("SpokeClusterCreatingController", recorder)
+}
+
+func (c *spokeClusterCreatingController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
deads2k, line 46: update this method so that it does 

```
get the expected instance with a lister
if it exists, return
if it doesn't, create
return any error
```

deads2k: you can do a live get if needed instead of the lister (that can be a little difficult to wire).

if you do this, then I think you eliminate your bools and special cases and this simply runs all the time, idle mostly.

deads2k:
improved.  A few more followup followups, but I'll tag

/lgtm
/approve

pkg/spoke/spokecluster/creating_controller.go
skeeey: I rewrite this part with get/create, and because the bootstrap config may be timeout, I cannot keep this controller always running

PR #17, opened 2020-04-30
rewrite spoke cluster creating controller

skeeey:


pmorie:
looks good.

/approve

pkg/spoke/spokeagent.go
```diff
@@ -90,6 +92,39 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
...
 
+	// load bootstrap client config
+	bootstrapClientConfig, err := clientcmd.BuildConfigFromFlags("", o.BootstrapKubeconfig)
+	if err != nil {
+		return fmt.Errorf("unable to load bootstrap kubeconfig from file %q: %w", o.BootstrapKubeconfig, err)
+	}
+
+	// start a SpokeClusterCreatingController to make sure there is a spoke cluster on hub cluster
+	// if the bootstrap client config is valid
+	if isBootstrapClientConfigValid(bootstrapClientConfig) {
```
deads2k: let's just require it by not testing this.  We can require well-formedness

pkg/spoke/spokeagent.go
```diff
@@ -90,6 +92,39 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
...
+		bootstrapClusterClient, err := clusterv1client.NewForConfig(bootstrapClientConfig)
+		if err != nil {
+			return err
+		}
+
+		caBundle := controllerContext.KubeConfig.CAData
```
deads2k: factor into a method please.

pkg/spoke/spokecluster/creating_controller.go
```diff
@@ -28,50 +27,50 @@ func NewSpokeClusterCreatingController(
...
-	spokeCluster := &clusterv1.SpokeCluster{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: c.clusterName,
-		},
-		Spec: clusterv1.SpokeClusterSpec{
-			SpokeClientConfig: clusterv1.ClientConfig{
-				URL:      c.spokeExternalServerUrl,
-				CABundle: c.spokeCABundle,
+	_, err := c.hubClusterClient.ClusterV1().SpokeClusters().Get(ctx, c.clusterName, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
```
deads2k:
```
switch
case unauthorized
    v4
case forbidden && error contains anonymous
    v4
case isNotFound
    fall to below
case err == nil
    return
case err != nil
    return err
```

Then this block isn't nested

pkg/spoke/spokeagent.go
skeeey: done

pkg/spoke/spokecluster/creating_controller.go
skeeey: done

pkg/spoke/spokeagent.go
skeeey: remove the testing code

pkg/spoke/spokecluster/creating_controller.go
```diff
@@ -28,24 +31,31 @@ func NewSpokeClusterCreatingController(
...
 		clusterName:            clusterName,
 		spokeExternalServerUrl: spokeExternalServerUrl,
 		spokeCABundle:          spokeCABundle,
-		hasSpokeClusterFunc:    hasSpokeClusterFunc,
 		hubClusterClient:       hubClusterClient,
 	}
 	return factory.New().
 		WithSync(c.sync).
-		ResyncEvery(5*time.Minute).
+		ResyncEvery(60*time.Minute).
```
deads2k, line 43: we may have to tighten this up later.  Don't be surprised at that outcome, but its good to start.

PR #18, opened 2020-04-30
enable integration test

skeeey:


test/integration-test.mk
```diff
@@ -0,0 +1,31 @@
+TEST_TMP :=/tmp
+
+export KUBEBUILDER_ASSETS ?=$(TEST_TMP)/kubebuilder/bin
+
+K8S_VERSION ?=1.16.4
+KB_TOOLS_ARCHIVE_NAME :=kubebuilder-tools-$(K8S_VERSION)-$(GOHOSTOS)-$(GOHOSTARCH).tar.gz
+KB_TOOLS_ARCHIVE_PATH := $(TEST_TMP)/$(KB_TOOLS_ARCHIVE_NAME)
+
+ensure-kubebuilder-tools:
```
pmorie, line 10: We should vendor things we are going to use in this way instead of downloading them, I think.

test/integration/integration_suite_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	err = clusterv1.AddToScheme(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
+	Expect(err).ToNot(HaveOccurred())
+	Expect(k8sClient).ToNot(BeNil())
+	close(done)
+}, 60)
+
```
pmorie, line 57: Some initial essential use cases to hit:

- spoke registration agent creates CSR, hub authorizes, spoke agent creates hub kubeconfig and connects back to hub for successful join
- agent recovery from invalid hub kubeconfig
- agent recovery from invalid bootstrap kubeconfig

test/integration-test.mk
pmorie: Hrm, the kubebuilder tools here are providing the API server binaries?

pmorie (commented):
Promising start, left a note about scenarios to cover.

test/integration-test.mk
skeeey: yes, it provides the api server and etcd binaries, we need them to start a api server

test/integration/integration_suite_test.go
skeeey: yes, they are useful hits, I add them to as comments firstly

test/integration/integration_suite_test.go
```diff
@@ -0,0 +1,62 @@
+package integration_test
+
+import (
+	"path/filepath"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
deads2k, line 8: keep the gomega alias please. We do this in many packages in kube

test/integration/integration_suite_test.go
```diff
@@ -0,0 +1,62 @@
+package integration_test
+
+import (
+	"path/filepath"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
```
deads2k, line 7: keep the ginkgo alias please

deads2k:
I'm ok to let this play out. Please let the aliases through.  Most packages I remember in kube use the aliases and it helps expectations when you're reading.

PR #20, opened 2020-05-06
Add deploy manifest

qiujian16:
This is to make installing registration easier

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,27 @@ You can reach the maintainers of this project at:
 
 Check the [Development Doc](docs/development.md) for how to contribute to the repo.
 
+## How to Deploy
+### Deploy Hub
+1. Run `make deploy-hub`
+
+### Deploy Spoke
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -23,6 +23,18 @@ You can reach the maintainers of this project at:
...
 Check the [Development Doc](docs/development.md) for how to contribute to the repo.
 
+## How to Deploy
+### Deploy Hub
+1. Run `make deploy-hub`
+
+### Deploy Spoke
+1. Create a bootstrap secret in the namespace of `open-cluster-management` with the following command:
+
+    ```sh
```
pmorie, line 42: We need to decode the secret and replace the URL with the kubernetes service cluster IP before `make deploy-spoke`.

I wouldn't mind having a make target in the future that makes the secret.

pmorie:
I was able to get the hub and spoke deployed following these instructions

/approve

PR #24, opened 2020-05-11
Add a name field selector to spoke cluster informer factory to solve agent cannot list spokeclusters problem

skeeey:
After hub cluster admin accept the spoke cluster, there will be errors in agent

```
E0511 06:45:49.129854       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.2/tools/cache/reflector.go:125: Failed to list *v1.SpokeCluster: spokeclusters.cluster.open-cluster-management.io is forbidden: User "system:open-cluster-management:spokecluster1:ghnl5" cannot list resource "spokeclusters" in API group "cluster.open-cluster-management.io" at the cluster scope
```
the root cause is `resourceNames` cannot be apply with list/watch rule

### DIFF
--- a/pkg/spoke/spokeagent.go
+++ b/pkg/spoke/spokeagent.go
@@ -184,7 +184,7 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
 	}
 
 	hubKubeInformerFactory := informers.NewSharedInformerFactory(hubKubeClient, 10*time.Minute)
-	// create an informer factory with cluster name field selector because we just need to handle the current spoke cluster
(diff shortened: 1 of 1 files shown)


deads2k:
/hold

You're looking for a construct like this to list/watch just the element you want: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/authentication/request/headerrequest/requestheader_controller.go#L111-L113

Doing that causes the authorizer to recognize it as a single element listwatch

skeeey:
@deads2k  I use `NewSharedInformerFactoryWithOptions` to replace the current `SharedInformerFactory` for spoke cluster

PR #26, opened 2020-05-11
update spoke-external-server-urls flag

skeeey:
Update `spoke-external-server-url` to `spoke-external-server-urls` as an string array flag


deads2k:
The field is optional in the spec because we wanted the spec creatable from either the hub or the spoke side.  I'm supportive of the idea of having clusters that are not externally accessible, but the correct order to do that is to first update the API indicating why this field could be empty and considering whether we should make the URLs an array to handle rotation or multiple potential external names.

/hold

holding for API to be updated before allowing this change.

skeeey:
OK, after API update finished, I will adjust this PR

pkg/spoke/spokecluster/creating_controller_test.go
```diff
@@ -69,14 +69,20 @@ func TestCreateSpokeCluster(t *testing.T) {
 
 func assertSpokeExternalServerUrl(t *testing.T, actual runtime.Object, expected string) {
 	spokeCluster := actual.(*clusterv1.SpokeCluster)
-	if spokeCluster.Spec.SpokeClientConfig.URL != expected {
-		t.Errorf("expected %q error, but got %q", expected, spokeCluster.Spec.SpokeClientConfig.URL)
+	if len(spokeCluster.Spec.SpokeClientConfigs) != 1 {
+		t.Errorf("expected one spoke client config, but got %v", spokeCluster.Spec.SpokeClientConfigs)
+	}
+	if spokeCluster.Spec.SpokeClientConfigs[0].URL != expected {
```
pmorie, line 75: Test the CA bundle data too (ok for follow-up)

deads2k:
squash before merge please
/lgtm

Also, can you point me at the bit updating this field?  The create uses different creds and that small controller is create, but something needs to be able to update it, right?

skeeey:
> Also, can you point me at the bit updating this field? The create uses different creds and that small controller is create, but something needs to be able to update it, right?

I do not implement the updating yet, because this filed may be updated by both hub and spoke
so the case is more complex.

I think we probably should not allow spoke cluster to update any field on spokeclustere.Spec after it is created

deads2k:
> I do not implement the updating yet, because this filed may be updated by both hub and spoke
> so the case is more complex.

hmmm.. We should talk about that.  It's owned by one or the other, but not both


/hold cancel
/lgtm

skeeey:
@deads2k , how about we do not allow spoke agent to update any field on `SpokeCluster.Spec` after the `SpokeCluster` is created? we just allow hub update the `SpokeCluster.Spec`, if this make sense, I will remove the `update` permission for spoke agent, currently spoke agent has below permissions

```
# Allow spoke agent to get/list/update/watch its owner spoke cluster
- apiGroups: ["cluster.open-cluster-management.io"]
  resources: ["spokeclusters"]
  resourceNames: ["{{ .SpokeClusterName }}"]
  verbs: ["get", "list", "update", "watch"]
```

PR #28, opened 2020-05-14
Add rules for agent

qiujian16:
1. manifestwork rules so work agent can get work
2. events rules for event recorder

### DIFF
--- a/pkg/hub/spokecluster/manifests/spokecluster-role.yaml
+++ b/pkg/hub/spokecluster/manifests/spokecluster-role.yaml
@@ -1,7 +1,0 @@
-apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
-metadata:
-  name: {{ .SpokeClusterName }}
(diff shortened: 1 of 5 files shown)


pkg/hub/spokecluster/manifests/spokecluster-role.yaml
```diff
@@ -3,5 +3,16 @@ kind: Role
 metadata:
   name: {{ .SpokeClusterName }}
```
deads2k: I think we're going to need names that distinguish
1. spoke-operator
2. spoke-registration
3. spoke-work

I'm having trouble keeping track and we don't have that many yet

pmorie: Agree.

I'm also not sure if this is the right place for these rules to live, but I think we can go with it for now.

qiujian16: i will change the name of these roles so it would be easier to distinguish

PR #31, opened 2020-05-18
Fix a couple missed kustomize invocations

pmorie:
...to use the installed kustomize.


deads2k:
Oddly opinionated, but we can try

/lgtm
/approve

PR #35, opened 2020-05-21
Add e2e binary to registration image

pmorie:
We may put this in a distinct image, ultimately, but the path of least resistance is to put it into the main registration image for now.


Dockerfile
```diff
@@ -4,7 +4,10 @@ COPY . .
 ENV GO_PACKAGE github.com/open-cluster-management/registration
 
 RUN make build --warn-undefined-variables
+RUN go test -c ./test/e2e
```
deads2k: nit: my preference is `make build-test`

deads2k:
just so it won't wait to merge

PR #37, opened 2020-05-27
add webhook e2e test cases

skeeey:


deads2k:
This looks like a good start. Did I just miss the positive test (user with permission updates the hubAccepted)?  if not, can you make it a follow-on?

/lgtm

skeeey:
> This looks like a good start. Did I just miss the positive test (user with permission updates the hubAccepted)? if not, can you make it a follow-on?

I think the `loopback registration` test already includes the positive test, we enabled the webhook with that test, in that test, the user who has the permission updates the hubAccepted to true

PR #44, opened 2020-06-15
add heartbeat e2e test case

skeeey:


test/e2e/heartbeat_test.go
```diff
@@ -0,0 +1,271 @@
...
+			lease, err := hubClient.CoordinationV1().Leases(clusterName).Get(context.TODO(), leaseName, metav1.GetOptions{})
+			if err != nil {
+				return false, err
+			}
+			lastRenewTime = lease.Spec.RenewTime
+			return true, nil
+		})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+		ginkgo.By(fmt.Sprintf("Make sure ManagedCluster lease %q is updated", leaseName))
```
pmorie: Why not just tack this onto the loopback test? That test can just be extended with the lease logic to become more thorough. WDYT?

skeeey: yes, it make sense, I will add this to loopback test, the loopback will test all positive cases

PR #46, opened 2020-06-18
fix fails when retry on conflict

zhiweiyin318:
fix conflict fails in e2e tests.
need to get the resource and then update the resource in RetryOnConflict


qiujian16:
good catch!

pmorie:
Very nice

/approve
/lgtm

PR #47, opened 2020-06-22
[wip] add CriticalResources

deads2k:
from https://github.com/openshift/enhancements/pull/380

```bash
make
make images
# push image
make deploy-criticalresource
# be sure you're using the image you created and pushed in the deployment
oc create -f example/criticalresource

[deads@deads-02 registration]$ oc delete deployment/protected-deployment
Error from server: admission webhook "criticalresourcesadmissionvalidators.admission.apiserver.openshift.io" denied the request: 1 instances of apiserver.openshift.io/v1, Resource=finalizerresources contain finalizer

[deads@deads-02 registration]$ oc delete criticalresource/protect-deployment
Error from server: admission webhook "criticalresourcesadmissionvalidators.admission.apiserver.openshift.io" denied the request: apps/v1, Resource=deployments named "protected-deployment" in namespace/openshift-criticalresource still exists

# remove finalizer
[deads@deads-02 registration]$ oc edit finalizerresource
finalizerresource.apiserver.openshift.io/openshift-criticalresource edited

[deads@deads-02 registration]$ oc delete deployment/protected-deployment
deployment.apps "protected-deployment" deleted
[deads@deads-02 registration]$ oc delete criticalresource/protect-deployment
criticalresource.apiserver.openshift.io "protect-deployment" deleted
```

There are overly broad permissions and unit tests are needed before merge, but I think you can see the broad strokes.  If we tightly constrain the resources we all provider usage on, we can avoid having dynamic registration in the first iteration.


pkg/criticalresource-admissionwebhook/webhook.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+func (a *CriticalResourceAdmissionWebhook) validateSpecificResourceCriteriaMet(ctx context.Context, namespace string, criteria apiserverv1.CriticalResourceCriteria) error {
+	// TODO fix API to include version
+	gvr := schema.GroupVersionResource{
+		Group:    criteria.SpecificResource.Group,
+		Version:  "v1",
+		Resource: criteria.SpecificResource.Resource,
+	}
+	_, err := a.dynamicClient.Resource(gvr).Namespace(namespace).Get(ctx, criteria.SpecificResource.Name, metav1.GetOptions{})
```
qiujian16: i think we have to check if the related resource is namespace or cluster scoped. I tested with
```
apiVersion: apiserver.openshift.io/v1
kind: CriticalResource
metadata:
  name: protect-deployment
spec:
  provider:
    group: apps
    resource: deployments
    name: protected-deployment
  criteria:
    - type: SpecificResource
      specificResource:
        group: apiextensions.k8s.io
        resource: customresourcedefinitions
        name: finalizerresources.apiserver.openshift.io
```

The namespace needs to be  set as "", then it works. I think we could get it by discovery client or restmapper.

pkg/criticalresource-admissionwebhook/webhook.go
```diff
@@ -0,0 +1,292 @@
...
+	gr := schema.GroupResource{
+		Group:    request.Resource.Group,
+		Resource: request.Resource.Resource,
+	}
+	switch gr {
+	case apiserverv1.Resource("criticalresources"):
+		return a.validateCriticalResourceDelete(ctx, request)
+	case corev1.Resource("namespace"):
+		return a.validateNamespaceDelete(ctx, request)
+	case appsv1.Resource("deployments"):
```
qiujian16: should we also validate manifestworks?

deads2k: Added

pkg/criticalresource-admissionwebhook/webhook.go
deads2k: Added

deads2k:
I added support for manifestworks as providers and for cluster scoped resources as criteria.  If we like the shape, I will start adding tests later this week.  For now, I'll focus on clearing out some reviews to unblock parallel efforts.

deploy/criticalresource/apiservice.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+metadata:
+  name: v1.admission.apiserver.openshift.io
+spec:
+  group: admission.apiserver.openshift.io
+  version: v1
+  service:
+    name: criticalresource-admissionwebhook
+    namespace: openshift-criticalresource
```
pmorie, line 10: Do we need to introduce a `openshift-` namespace here? I'd prefer not to in this repo if we can possibly avoid it.

pmorie: How about open-cluster-management-criticalresource?

vendor/github.com/openshift/api/apiserver/v1/types.go
```diff
@@ -0,0 +1,83 @@
...
+type CriticalResource struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   CriticalResourceSpec   `json:"spec"`
+	Status CriticalResourceStatus `json:"status"`
+}
+
+type CriticalResourceSpec struct {
+	// provider indicates the resource which provides critical functionality.  It must be in the same namespace as the
```
pmorie, line 25: Just a little bit of bikeshedding, I think `resource` might be an easier name for folks to understand

deploy/criticalresource/apiservice.yaml
pmorie: I understand now why you used an `openshift-` namespace in this prototype, we should think carefully about whether we would ship this with that ns at this point, would the resource live in ocm/api, etc.

vendor/github.com/openshift/api/apiserver/v1/types.go
```diff
@@ -0,0 +1,83 @@
...
+	SpecificResourceType CriticalResourceCriteriaType = "SpecificResource"
+)
+
+type CriticalResourceCriteria struct {
+	Type             CriticalResourceCriteriaType              `json:"type"`
+	Finalizer        *FinalizerCriticalResourceCriteria        `json:"finalizer"`
+	SpecificResource *SpecificResourceCriticalResourceCriteria `json:"specificResource"`
+}
+
+type FinalizerCriticalResourceCriteria struct {
```
pmorie, line 60: I would like to see godoc for this and specificresource so we can debate their exact semantics.

vendor/github.com/openshift/api/apiserver/v1/types.go
sttts: +1

Provider is not intuitive. Without reading here you won't understand what a provider is for a critical resource. Just call it "resource", or even inline.

vendor/github.com/openshift/api/apiserver/v1/types.go
```diff
@@ -0,0 +1,83 @@
...
+	Status CriticalResourceStatus `json:"status"`
+}
+
+type CriticalResourceSpec struct {
+	// provider indicates the resource which provides critical functionality.  It must be in the same namespace as the
+	// critical resource.
+	// TODO if we need to protect cluster scoped resources, we will need to create a resource type.  We cannot reasonably
+	//  block deletion of a cluster scoped resource from within a namespace.
+	Provider CriticalResourceProvider   `json:"provider"`
+	Criteria []CriticalResourceCriteria `json:"criteria"`
```
sttts, line 30: criteria of what? This isn't intuitive either. You define deletion blockers here. Calls them with what they are.

PR #49, opened 2020-06-24
deleting managed cluster

skeeey:
Using clean up managed cluster manifest files instead of only deleting namespace, thus, we can make sure the clusterrole and role are deleted when failed to delete namespace

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -2,16 +2,44 @@ package helpers
 
 import (
 	"context"
+	"fmt"
 	"net/url"
+	"path/filepath"
 	"time"
 
 	clusterclientset "github.com/open-cluster-management/api/client/cluster/clientset/versioned"
 	clusterv1 "github.com/open-cluster-management/api/cluster/v1"
+	"github.com/open-cluster-management/registration/pkg/hub/managedcluster/bindata"
+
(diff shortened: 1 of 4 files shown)


jnpacker:
Bug Issue: https://github.com/open-cluster-management/backlog/issues/3059

pkg/helpers/helpers.go
```diff
@@ -128,3 +156,149 @@ func IsValidHTTPSURL(serverURL string) bool {
...
+			errs = append(errs, err)
+			continue
+		}
+		gvk := resourcehelper.GuessObjectGroupVersionKind(object)
+		recorder.Eventf(fmt.Sprintf("ManagedCluster%sDeleted", gvk.Kind), "Deleted %s", resourcehelper.FormatResourceForCLIWithNamespace(object))
+	}
+	return errorhelpers.NewMultiLineAggregate(errs)
+}
+
+// CleanUpGroupFromClusterRoleBindings search all clusterrolebings for managed cluster group and remove the subject entry
```
pmorie, line 206: typo: `clusterrolebings`

deads2k:
I see where your head is at.  I suspect you'll have more work to do, but this seems like a reasonable starting point.

/lgtm
/approve

PR #50, opened 2020-06-30
keeping namespace when deleting managed cluster

skeeey:


deads2k:
Clever.

I heard that the rcm-controller will watch for the removal of both the ClusterDeployment and ManagedCluster as the conditions to remove the namespace.  It would be nice to have that PR linked to this, because this leaves it dangling.

/lgtm
/approve

PR #53, opened 2020-07-02
Add details (metric) about Spoke cluster

bjoydeep:
This is a WIP PR. But since this could generate lots of discussions, wanted raise this and capture all the discussion through this PR.

Goal:
We aim to capture a metric about the Spoke cluster that joins the Hub in Prometheus exposition format as shown
```
//cluster_id = OCP ID of the Cluster (need to resolve for eks, etc)
//type = K8s Distribution, e.g. OCP, EKS, etc
//version = Distribution version
//cluster_infrastructure_provider = value "Type" from cluster_infrastructure_provider
//hub_id = cluster_id of hub server
//cluster_name =User Display Name of Cluster (defaults to Id if not provided)
var managedClusterMetric = prometheus.NewGaugeVec(prometheus.GaugeOpts{
	Name: "a_managed_cluster",
	Help: "Managed Cluster being managed by ACM Hub.",
}, []string{"cluster_name", "cluster_id", "type", "version", "cluster_infrastructure_provider", "hub_id"})
```
Highlights:
1. Metrics endpoint is created such that it can be scraped by Prometheus. For this reason we have exposed a new API endpoint in https://github.com/bjoydeep/registration/tree/master/pkg/hub/custommetrics. We also have been told by @deads2k and @qiujian16 that we should NOT expose this endpoint - but use the default 8443 port which is available for this and already exposes some other key metrics about the performance of the controller. Moving to 8443 will be WIP.
2. For the code to work the way it is, we need to add an environment variable  (ie we have fenced this off): https://github.com/bjoydeep/registration/blob/master/pkg/hub/manager.go#L88-#L102 . So unless the env variable is specifically added to the deployment yaml, this code will not get called.
3. We are working on adding e2e tests.
4. As of now, we can only gather the cluster-name of the spoke cluster (which may or may not be the OpenShift ClusterID for OpenShift clusters). We will propose to make other changes so that we can get the ClusterID for OpenShift and  *KS clusters in a consistent manner. But we can do that elsewhere. We have also captured the ClusterID for the Spoke cluster using the OpenShift ClusterVersionOperator
5. We are trying to see what metrics are gathered at the default 8443 port of the registration controller POD and running into security issues - which we will try to work through.


bjoydeep:
Quick update: @qiujian16 and I played a little bit with what is available in the registration-controller POD at port 8443. Using the token of the secret mounted on the POD, and granting the POD service account a cluster-admin role, we could see all metrics on https://localhost:8443/metrics. Should we not formally expose this service and a ServiceMonitor CR (much like the operator sdk does)?

deads2k:
> and granting the POD service account a cluster-admin role

You should be able to bound that to a role that only allows access to /metrics.

deploy/hub/hub_controller_clusterrole.yaml
```diff
@@ -38,3 +38,7 @@ rules:
 - apiGroups: ["register.open-cluster-management.io"]
   resources: ["managedclusters/accept"]
   verbs: ["update"]
+# Allow hub to read cluster version
```
deads2k: can this comment be updated to say why it needs this access?

pkg/hub/custommetrics/custommetrics.go
```diff
@@ -0,0 +1,198 @@
...
+//cluster_infrastructure_provider = value "Type" from cluster_infrastructure_provider
+//hub_id = cluster_id of hub server
+//cluster_name =User Display Name of Cluster (defaults to Id if not provided)
+var managedClusterMetric = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+	Name: "a_managed_cluster",
+	Help: "Managed Cluster being managed by ACM Hub.",
+}, []string{"cluster_name", "cluster_id", "type", "version", "cluster_infrastructure_provider", "hub_id"})
+
+func getDynClient() (dynamic.Interface, error) {
+	config, err := rest.InClusterConfig()
```
deads2k: this makes it impossible to de-containerize your app.  The construction of this binary current builds a rest.Config.  I suggest passing that in some way.

pkg/hub/custommetrics/custommetrics.go
```diff
@@ -0,0 +1,198 @@
...
+		panic(errCv.Error())
+	}
+	cv := &ocinfrav1.ClusterVersion{}
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(cvObj.UnstructuredContent(), &cv)
+	if err != nil {
+		klog.Fatalf("Error unmarshal cluster version object%v \n", err)
+		panic(errCv.Error())
+	}
+	hubID = string(cv.Spec.ClusterID)
+	klog.Infof("Hub id of this ACM Server is " + hubID)
```
deads2k: don't log from utility libraries.  Especially if you're going to report it later.

pkg/hub/custommetrics/custommetrics.go
```diff
@@ -0,0 +1,198 @@
...
+
+}
+
+func addCluster(obj interface{}) {
+	klog.Info("Adding Managed Clusters ####################")
```
deads2k: no eye catchers and loglevel needs adjusting.

pkg/hub/custommetrics/custommetrics.go
```diff
@@ -0,0 +1,198 @@
...
+	klog.Infof("Managed Cluster name being removed: %s \n", managedCluster.GetName())
+
+	//TODO:
+	//get the actual values as mentioned here:
+	//https://github.com/open-cluster-management/perf-analysis/blob/master/Big%20Picture.md#acm-20-telemetry-data
+	managedClusterMetric.WithLabelValues(managedCluster.GetName(), "cluster_id", "type", "version", managedCluster.GetLabels()["cloud"], hubID).Set(0)
+
+}
+
+func fetchTestData(c dynamic.Interface) {
```
deads2k: needs to be cleaned up before merge.

pkg/hub/custommetrics/custommetrics.go
```diff
@@ -0,0 +1,198 @@
...
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "Hello, %q, how are you ?\n", html.EscapeString(r.URL.Path))
+
+	})
+
+	http.Handle("/metrics", promhttp.HandlerFor(r, promhttp.HandlerOpts{}))
+
+	klog.Infof("Custom Metric Service started on port: " + fmt.Sprint(customMetricsPort))
+	klog.Fatal(http.ListenAndServe(":"+fmt.Sprint(customMetricsPort), nil))
```
deads2k: this looks inappropriate to me.  Just some testing code?  it's a little difficult to tell if this is a WIP or not.

You have a comment mentioning that 8443 is serving metrics. Is it via this or via the library-go server launch (which also holds healthz)

pkg/hub/manager.go
```diff
@@ -81,6 +85,22 @@ func RunControllerManager(ctx context.Context, controllerContext *controllercmd.
 	go leaseController.Run(ctx, 1)
 	go rbacFinalizerController.Run(ctx, 1)
 
+	//Add Custom Metrics
+	//make sure its a go func call else it will block
+	enableMetric := false
+	val, exists := os.LookupEnv("METRIC_ENABLE")
```
deads2k: metrics should always be enabled.

pkg/hub/custommetrics/custommetrics.go
```diff
@@ -0,0 +1,198 @@
...
+
+	hubID = ""
+)
+
+//cluster_id = OCP ID of the Cluster (need to resolve for eks, etc)
+//type = K8s Distribution, e.g. OCP, EKS, etc
+//version = Distribution version
+//cluster_infrastructure_provider = value "Type" from cluster_infrastructure_provider
+//hub_id = cluster_id of hub server
+//cluster_name =User Display Name of Cluster (defaults to Id if not provided)
```
deads2k, line 49: I don't think we normally provide these.

pkg/hub/custommetrics/custommetrics.go
```diff
@@ -0,0 +1,198 @@
...
+	/* cluster := &clusterv1.ManagedCluster{}
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(mc.UnstructuredContent(), &cluster)
+	if err != nil {
+		klog.Fatalf("Error unmarshal managed cluster object%v \n", err)
+	} */
+
+	//TODO:
+	//get the actual values as mentioned here:
+	//https://github.com/open-cluster-management/perf-analysis/blob/master/Big%20Picture.md#acm-20-telemetry-data
+	managedClusterMetric.WithLabelValues(managedCluster.GetName(), "cluster_id", "type", "version", managedCluster.GetLabels()["cloud"], hubID).Set(1)
```
deads2k, line 95: Let's get all the mechanics in place and merged with this line and the other set under 
```
if false{
  // do it
```

That way we can get everything else merged

deads2k:
if we do this (https://github.com/open-cluster-management/registration/pull/53#pullrequestreview-446667330) , then we'll be down to arguing about exactly what can set as the labelled stuff.  I think that puts you in a very good spot.

bjoydeep:
As advised by @deads2k
Falling back on `/metrics` port already opened by the registration controller. 
And therefore using the `legacyregistry`. 
And addressed some of the other comments by David.

Working on adding e2e test.

deploy/hub/hub_controller_clusterrole.yaml
bjoydeep: added comments David.

pkg/hub/custommetrics/custommetrics.go
bjoydeep: Tried to follow the example you pointed to.

pkg/hub/custommetrics/custommetrics.go
bjoydeep: agreed - removed

pkg/hub/custommetrics/custommetrics.go
bjoydeep: removed the eye catcher. Wondering on the level here - will it very noisy at info. Are we expecting clusters to come and go very frequently?

pkg/hub/custommetrics/custommetrics.go
bjoydeep: I am hoping you meant a proper name and not fetchTestData. If so, it has the correct name now.

pkg/hub/custommetrics/custommetrics.go
bjoydeep: this is gone as we are using the endpoint already created.

pkg/hub/manager.go
bjoydeep: correct - removed.

pkg/hub/custommetrics/custommetrics.go
bjoydeep: Did not exactly follow this one. Let me know please.

qiujian16:
close since the PR is stale

PR #54, opened 2020-07-03
try healthz api when readyz api is forbidden

skeeey:


skeeey:
see problem https://github.com/open-cluster-management/backlog/issues/3186

PR #55, opened 2020-07-03
rename clusterrole and clusterrolebinding

zhiweiyin318:
related issue: https://github.com/open-cluster-management/backlog/issues/3025


pmorie:
Weird, looks like the unit test container was killed...

/retest

PR #56, opened 2020-07-06
Using specific role instead of cluster admin to deploy registration agent

skeeey:
Using specific role instead of cluster admin to deploy registration agent, so make sure the e2e test more credible

### DIFF
--- a/deploy/spoke/clusterrole.yaml
+++ b/deploy/spoke/clusterrole.yaml
@@ -1,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: open-cluster-management:spoke
+rules:
+- apiGroups: [""]
+  resources: ["nodes", "configmaps", "secrets"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["authorization.k8s.io"]
+  resources: ["subjectaccessreviews"]
+  verbs: ["create"]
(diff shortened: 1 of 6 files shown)


deploy/spoke/role.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: open-cluster-management:spoke
```
pmorie: Let's think carefully about using the name `spoke` here. `agent` might be a better alternative. Do you have a preference on this detail, @mdelder ?

pmorie (commented):
Good change, let's just make sure we have the right name

deploy/spoke/role.yaml
skeeey: how about `registration-agent`? I think this more specific

mdelder:
I agree with @pmorie on this one. I would be vote for either registration-agent or klusterlet as the name for the role.

skeeey:
using `registration-agent` first

deploy/spoke/clusterrole.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: open-cluster-management:spoke
```
deads2k, line 4: I think we re-branded "managed-cluster" so you may want to update this.

PR #57, opened 2020-07-10
refactor ut code

skeeey:
1. reorganize ut common functions
2. improve test coverage

### DIFF
--- a/pkg/helpers/testing/assertion.go
+++ b/pkg/helpers/testing/assertion.go
@@ -1,0 +1,193 @@
+package testing
+
+import (
+	"reflect"
+	"testing"
+
+	clusterv1 "github.com/open-cluster-management/api/cluster/v1"
+
+	authorizationv1 "k8s.io/api/authorization/v1"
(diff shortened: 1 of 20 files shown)


pkg/helpers/testing/assertion.go
```diff
@@ -0,0 +1,174 @@
...
+		t.Errorf("expected %q error, but got %q", expectedErr, actual.Error())
+		return
+	}
+	if len(expectedErr) == 0 && actual != nil {
+		t.Errorf("unexpected err: %v", actual)
+		return
+	}
+}
+
+func AssertActions(t *testing.T, actualActions []clienttesting.Action, expectedActions ...string) {
```
deads2k: I recommend name change and doc. `expectedVerbs`?

deads2k:
something broke in namespaces, but I like the idea

/approve
/assign @qiujian16

pkg/helpers/testing/assertion.go
skeeey: renamed and add the doc for each  assert method

PR #66, opened 2020-08-24
Add controller for ManagedClusterSet

elgnay:


test/integration/managedclusterset_test.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+			for _, condition := range managedClusterSet.Status.Conditions {
+				if condition.Type != clusterv1alpha1.ManagedClusterSetConditionEmpty {
+					continue
+				}
+				if condition.Status != metav1.ConditionTrue {
+					return false
+				}
+				if condition.Reason != "NoClusterMatched" {
+					//if condition.Reason != "ClustersSelected" {
```
qiujian16: does this line need removed?

elgnay: Removed.

pkg/hub/managedclusterset/controller.go
```diff
@@ -0,0 +1,265 @@
...
+		WithSync(c.sync).
+		ToController("ManagedClusterSetController", recorder)
+}
+
+func (c *managedClusterSetController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	key := syncCtx.QueueKey()
+
+	// if the sync is triggered by change of ManagedCluster, reconcile all cluster sets
+	if key == "key" {
+		if err := c.syncAllClusterSets(ctx); err != nil {
```
deads2k, line 66: make a TODO to better than this.  You only need to recheck the ones that contain the managed cluster in question, right?  You can build this with a reverse index.

pkg/hub/managedclusterset/controller.go
```diff
@@ -0,0 +1,265 @@
...
+	clusterSetInformer clusterinformerv1alpha1.ManagedClusterSetInformer,
+	recorder events.Recorder) factory.Controller {
+	c := &managedClusterSetController{
+		clusterClient:    clusterClient,
+		clusterLister:    clusterInformer.Lister(),
+		clusterSetLister: clusterSetInformer.Lister(),
+		eventRecorder:    recorder.WithComponentSuffix("managed-cluster-set-controller"),
+	}
+	return factory.New().
+		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
```
deads2k, line 47: Looks like a missing part of the library.   You need to produce a `[]string` based on one object.  You ought to create  PR fixing it in library-go.  And add a TODO here

pkg/hub/managedclusterset/controller.go
deads2k: > make a TODO to better than this. You only need to recheck the ones that contain the managed cluster in question, right? You can build this with a reverse index.

Even if you have to list everything, the workqueue behavior will help reduce storms if you compute the individual names.

pkg/hub/managedclusterset/controller.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	clusterSet = clusterSet.DeepCopy()
+	clusterSet.Status.Conditions = newConditions
+	_, err := c.clusterClient.ClusterV1alpha1().ManagedClusterSets().UpdateStatus(ctx, clusterSet, metav1.UpdateOptions{})
+	return err
+}
+
+// unique returns a new slice with the unique items.
+func unique(items []string) (uniqueItems []string) {
```
deads2k: k8s.io/util sets.String I think.

pkg/hub/managedclusterset/controller.go
```diff
@@ -0,0 +1,265 @@
...
+	for _, selector := range selectors {
+		clusters, err := c.selectClusters(selector)
+		if err != nil {
+			return err
+		}
+
+		selectedClusters = append(selectedClusters, clusters...)
+	}
+
+	selectedClusters = unique(selectedClusters)
```
deads2k: I don't see this actually doing anything.  Your plan is to add that later?  That has some impact on whether or not you fail open or closed if a particular expression doesn't work.

pkg/hub/managedclusterset/controller.go
```diff
@@ -0,0 +1,265 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to list ManagedClusters with label selector: %v, %w", selector.LabelSelector, err)
+		}
+		for _, cluster := range clusters {
+			clusterNames = append(clusterNames, cluster.Name)
+		}
+		return clusterNames, nil
+	default:
+		// select all clusters if neither ClusterNames nor LabelSelector is specified
+		clusters, err := c.clusterLister.List(labels.Everything())
```
deads2k: The admission plugin needs to correctly handle this right?

deads2k:
looks reasonable.  My comments don't need to block.

/approve

pkg/hub/managedclusterset/controller.go
pmorie: As currently written, with sufficient number of clusters this implementation will not finish reconciling before the refresh interval is exhausted.

pkg/hub/managedclusterset/controller.go
pmorie: @deads2k I think you might want to define "fail open" and "fail closed" here.

pkg/hub/managedclusterset/controller.go
pmorie: Hrm, I am not sure if I would prefer an empty spec to mean "everything" or "empty". I would probably prefer "empty", which I also think is a valid construct. I feel like if you want to make a set that matches everything, you should have to say so explicitly, since I would view this as a relatively infrequent case.

pkg/hub/managedclusterset/controller_test.go
```diff
@@ -0,0 +1,349 @@
...
+			continue
+		}
+		return true
+	}
+	return false
+}
+
+// elementsMatch compares x is equal to y ignoring the order of the elements.
+// It assumes that there is no duplicated item in the input slices
+func elementsMatch(xs, ys []string) bool {
```
pmorie: You'll be able to take this out once you use stringset.

pkg/hub/managedclusterset/controller.go
elgnay: TODO is added and will deal with it in a follow-up PR.

pkg/hub/managedclusterset/controller.go
elgnay: TODO added.

pkg/hub/managedclusterset/controller.go
elgnay: Code refactored with sets.String.

pkg/hub/managedclusterset/controller.go
elgnay: Updated the code to fail closed.

pkg/hub/managedclusterset/controller.go
elgnay: Updated to select no cluster if neither ClusterNames nor LabelSelector is specified. And I'll create an another PR to update comments on CRD.

pkg/hub/managedclusterset/controller_test.go
elgnay: Code refactored with sets.String.

PR #67, opened 2020-08-24
Add validating webhook for ManagedClusterSetBinding

elgnay:


pkg/webhook/clustersetbinding/validating_webhook.go
```diff
@@ -0,0 +1,166 @@
...
+		status.Allowed = false
+		status.Result = &metav1.Status{
+			Status: metav1.StatusFailure, Code: http.StatusBadRequest, Reason: metav1.StatusReasonBadRequest,
+			Message: "Spec.ClusterSet is required",
+		}
+		return status
+	}
+
+	// check if the target cluster set exists
+	_, err := a.clusterClient.ClusterV1alpha1().ManagedClusterSets().Get(context.TODO(), binding.Spec.ClusterSet, metav1.GetOptions{})
```
qiujian16: should we allow creating a clustersetbinding ref to a non-exist clusterset.

qiujian16 (commented):
i think we need a e2e test to ensure to test webhooks.

pkg/webhook/clustersetbinding/validating_webhook.go
elgnay: Removed the existence checking logic.

elgnay:
> i think we need a e2e test to ensure to test webhooks.

Added.

pkg/webhook/clustersetbinding/validating_webhook.go
```diff
@@ -0,0 +1,145 @@
...
+		status.Allowed = false
+		status.Result = &metav1.Status{
+			Status: metav1.StatusFailure, Code: http.StatusBadRequest, Reason: metav1.StatusReasonBadRequest,
+			Message: err.Error(),
+		}
+		return status
+	}
+
+	// check if target cluster set is not specified
+	if len(binding.Spec.ClusterSet) == 0 {
```
deads2k: you can do this much in CRD schemas.  Get up with mdame in slack or #forum-api-review

pkg/webhook/clustersetbinding/validating_webhook.go
```diff
@@ -0,0 +1,145 @@
...
+		Spec: authorizationv1.SubjectAccessReviewSpec{
+			User:   userInfo.Username,
+			UID:    userInfo.UID,
+			Groups: userInfo.Groups,
+			Extra:  extra,
+			ResourceAttributes: &authorizationv1.ResourceAttributes{
+				Group:       "cluster.open-cluster-management.io",
+				Resource:    "managedclustersets",
+				Subresource: "bind",
+				Verb:        "get",
```
deads2k: `create` feels more natural

pkg/webhook/clustersetbinding/validating_webhook.go
```diff
@@ -0,0 +1,145 @@
...
+func (a *ManagedClusterSetBindingValidatingAdmissionHook) allowBindingToClusterSet(clusterSetName string, userInfo authenticationv1.UserInfo) *admissionv1beta1.AdmissionResponse {
+	status := &admissionv1beta1.AdmissionResponse{}
+
+	extra := make(map[string]authorizationv1.ExtraValue)
+	for k, v := range userInfo.Extra {
+		extra[k] = authorizationv1.ExtraValue(v)
+	}
+
+	sar := &authorizationv1.SubjectAccessReview{
+		Spec: authorizationv1.SubjectAccessReviewSpec{
```
deads2k, line 90: this should be documented in your api godoc.  apologies if you did this after I reviewed before.

pkg/webhook/clustersetbinding/validating_webhook.go
```diff
@@ -0,0 +1,145 @@
...
+			ResourceAttributes: &authorizationv1.ResourceAttributes{
+				Group:       "cluster.open-cluster-management.io",
+				Resource:    "managedclustersets",
+				Subresource: "bind",
+				Verb:        "get",
+				Name:        clusterSetName,
+			},
+		},
+	}
+	sar, err := a.kubeClient.AuthorizationV1().SubjectAccessReviews().Create(context.TODO(), sar, metav1.CreateOptions{})
```
deads2k, line 104: seems like this check should only be done if the value of clusterset is changing. if you're just adding an annotation or something, is there a reason to restrict access like this?

pkg/webhook/clustersetbinding/validating_webhook.go
```diff
@@ -0,0 +1,145 @@
...
+	// check if target cluster set is not specified
+	if len(binding.Spec.ClusterSet) == 0 {
+		status.Allowed = false
+		status.Result = &metav1.Status{
+			Status: metav1.StatusFailure, Code: http.StatusBadRequest, Reason: metav1.StatusReasonBadRequest,
+			Message: "Spec.ClusterSet is required",
+		}
+		return status
+	}
+
```
deads2k: how do you feel about forcing the names to match.  So that .metadata.name == .spec.clusterset?

pkg/webhook/clustersetbinding/validating_webhook.go
elgnay: Updated.

pkg/webhook/clustersetbinding/validating_webhook.go
elgnay: Doc is added

pkg/webhook/clustersetbinding/validating_webhook.go
elgnay: Removed the permission checking when updating cluster set binding.

pkg/webhook/clustersetbinding/validating_webhook.go
elgnay: Restriction is added to ensure cluster set binding and its target cluster set have the same name.

pkg/webhook/clustersetbinding/validating_webhook.go
elgnay: Removed the checking and added a validation rule on CRD.

pkg/webhook/clustersetbinding/validating_webhook.go
qiujian16: If we force the name to be matched, is .spec.clusterset still necessary? I think we can reconsider this as followup

PR #91, opened 2020-10-22
Enable metrics

itdove:


qiujian16:
this is no longer needed.

PR #92, opened 2020-11-02
Use eventfilterinformer

qiujian16:


pkg/spoke/hubclientcert/controller.go
```diff
@@ -103,34 +103,30 @@ func NewClientCertForHubController(
...
-			return irrelevantSecretKey
-		}, spokeSecretInformer.Informer()).
+			},
+			func(obj interface{}) bool {
+				accessor, err := meta.Accessor(obj)
+				if err != nil {
+					return false
+				}
+				// only enqueue when hub kubeconfig secret is changed
+				if accessor.GetNamespace() == hubKubeconfigSecretNamespace || accessor.GetName() == kubeconfigSecretName {
```
skeeey: I think this should be `&&`, right?

pkg/spoke/hubclientcert/secret_controller.go
```diff
@@ -47,27 +47,28 @@ func NewHubKubeconfigSecretController(
...
+			func(obj interface{}) bool {
 				accessor, err := meta.Accessor(obj)
 				if err != nil {
-					return ""
+					return false
 				}
-				if accessor.GetNamespace() == hubKubeconfigSecretNamespace && accessor.GetName() == hubKubeconfigSecretName {
-					return accessor.GetName()
+				// only enqueue when hub kubeconfig secret is changed
+				if accessor.GetNamespace() == hubKubeconfigSecretNamespace || accessor.GetName() == hubKubeconfigSecretName {
```
skeeey: I think this should be `&&`, right?

pkg/spoke/hubclientcert/controller.go
```diff
@@ -103,34 +103,30 @@ func NewClientCertForHubController(
...
-			if accessor.GetNamespace() == hubKubeconfigSecretNamespace && accessor.GetName() == kubeconfigSecretName {
+		WithFilteredEventsInformersQueueKeyFunc(
+			func(obj runtime.Object) string {
+				accessor, _ := meta.Accessor(obj)
 				return accessor.GetName()
-			}
-			return irrelevantSecretKey
-		}, spokeSecretInformer.Informer()).
+			},
+			func(obj interface{}) bool {
```
deads2k, line 111: I'm open to pushing this into a function in library-go for, `filterByInstance(ns, name) func` or some such.

PR #95, opened 2020-11-17
increase qps and burst

skeeey:


skeeey:
refer to https://github.com/open-cluster-management/backlog/issues/7072

pkg/hub/manager.go
```diff
@@ -18,21 +18,28 @@ import (
 	"github.com/open-cluster-management/registration/pkg/hub/rbacfinalizerdeletion"
 
 	kubeinformers "k8s.io/client-go/informers"
+	"k8s.io/client-go/rest"
 )
 
 // RunControllerManager starts the controllers on hub to manage spoke cluster registration.
 func RunControllerManager(ctx context.Context, controllerContext *controllercmd.ControllerContext) error {
-	kubeClient, err := kubernetes.NewForConfig(controllerContext.KubeConfig)
+	// Increase the QPS and Burst to enhance the ability of kube client to hanle requstes requests in concurrent
```
qiujian16: typo handle requests

pkg/hub/manager.go
```diff
@@ -18,21 +18,28 @@ import (
...
+	// TODO: these may be tuned by the flag of library-go when new a controller command in the future
```
qiujian16: change some wording here. Something like:

Use flag to change qps/burst

pkg/hub/manager.go
```diff
@@ -18,21 +18,28 @@ import (
...
+	kubeConfig := rest.CopyConfig(controllerContext.KubeConfig)
+	kubeConfig.QPS = 100.0
```
qiujian16: maybe we should  set this only when QPS in kubconfig is not set.

pkg/hub/manager.go
skeeey: corrected

pkg/hub/manager.go
skeeey: changed

pkg/hub/manager.go
skeeey: add a judgement for it

PR #96, opened 2020-11-24
RBAC escalation for granting per-cluster RBAC-rules to managedclusters

yue9944882:
the managedcluster-controller will be creating per-cluster role/rolebinding to the corresponding namespace which requires the RBAC permission escalation. this pull adds sufficient RBAC permission to the default manifests so that the automatic creation will work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ e2e.test
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 _output/
+.idea/
--- a/deploy/hub/managedcluster_escalation_clusterrole.yaml
+++ b/deploy/hub/managedcluster_escalation_clusterrole.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: open-cluster-management:hub:escalation
+rules:
+# Allow work agent to get/list/watch/update manifestworks
+- apiGroups: ["work.open-cluster-management.io"]
+  resources: ["manifestworks"]
+  verbs: ["get", "list", "watch", "update"]
+# Allow work agent to update the status of manifestwork
+- apiGroups: ["work.open-cluster-management.io"]
+  resources: ["manifestworks/status"]
+  verbs: ["patch", "update"]
--- a/deploy/hub/managedcluster_escalation_clusterrolebinding.yaml
+++ b/deploy/hub/managedcluster_escalation_clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: open-cluster-management:hub:escalation
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: open-cluster-management:hub:escalation
+subjects:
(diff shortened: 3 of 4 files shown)


deploy/hub/managedcluster_escalation_clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
...
+    resources: ["events"]
+    verbs: ["create", "patch", "update"]
+  # Allow work agent to get/list/watch/update manifestworks
+  - apiGroups: ["work.open-cluster-management.io"]
+    resources: ["manifestworks"]
+    verbs: ["get", "list", "watch", "update"]
+  # Allow work agent to update the status of manifestwork
+  - apiGroups: ["work.open-cluster-management.io"]
+    resources: ["manifestworks/status"]
+    verbs: ["patch", "update"]
```
qiujian16: EOF?

deploy/hub/managedcluster_escalation_clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: open-cluster-management:hub:escalation
+rules:
+  # Allow work agent to send event to hub
+  - apiGroups: ["", "events.k8s.io"]
```
qiujian16: events is declared already in https://github.com/open-cluster-management/registration/blob/master/deploy/hub/hub_controller_clusterrole.yaml#L15, where i think we could add group of events.k8s.io

qiujian16 (commented):
Thanks! just minor comments.

.gitignore
```diff
@@ -15,3 +15,4 @@ e2e.test
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 _output/
+.idea/
```
skeeey: Add an EOF, please

deploy/hub/managedcluster_escalation_clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: open-cluster-management:hub:escalation
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: open-cluster-management:hub:escalation
+subjects:
+  - kind: ServiceAccount
+    name: hub-sa
+    namespace: open-cluster-management
```
skeeey: EOF too, please

deploy/hub/managedcluster_escalation_clusterrole.yaml
yue9944882: addressed

deploy/hub/managedcluster_escalation_clusterrole.yaml
yue9944882: added "events.k8s.io" group and tightened verbs by removing "list" and "watch" which is not in use by the recorder

.gitignore
yue9944882: addressed.

deploy/hub/managedcluster_escalation_clusterrolebinding.yaml
yue9944882: addressed.

qiujian16:
/lgtm

Thanks!

PR #97, opened 2020-11-26
Add ClusterClaim controller

elgnay:

The task link: https://github.com/open-cluster-management/backlog/issues/7156


deploy/spoke/clusterrole.yaml
```diff
@@ -9,3 +9,7 @@ rules:
 - apiGroups: ["authorization.k8s.io"]
   resources: ["subjectaccessreviews"]
   verbs: ["create"]
+# Allow agent to list clusterclaims
+- apiGroups: ["cluster.open-cluster-management.io"]
+  resources: ["clusterclaims"]
+  verbs: ["get", "list", "watch"]
```
qiujian16: EOF

pkg/spoke/managedcluster/claim_controller.go
```diff
@@ -0,0 +1,146 @@
...
+			return accessor.GetName()
+		}, hubManagedClusterInformer.Informer()).
+		WithSync(c.sync).
+		ToController("ManagedClusterClaimController", recorder)
+}
+
+// sync maintains the cluster claims in status of the managed cluster on hub once it joins the hub.
+func (c managedClusterClaimController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	key := syncCtx.QueueKey()
+	if key != c.clusterName && key != factory.DefaultQueueKey {
```
qiujian16: do we need this check? We do not need this key, right?

deploy/spoke/clusterrole.yaml
elgnay: EOF added.

pkg/spoke/managedcluster/claim_controller.go
elgnay: You are right. We don't need the check because the informer has already had the same filter configured. And the key is unnecessary either.

deploy/hub/managedcluster.crd.yaml
```diff
@@ -3,13 +3,29 @@ kind: CustomResourceDefinition
 metadata:
   name: managedclusters.cluster.open-cluster-management.io
 spec:
+  additionalPrinterColumns:
```
pmorie, line 6: we didn't have these before? weird

pmorie (commented):
looking good

pkg/spoke/managedcluster/claim_controller.go
```diff
@@ -0,0 +1,140 @@
...
+// sync maintains the cluster claims in status of the managed cluster on hub once it joins the hub.
+func (c managedClusterClaimController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	managedCluster, err := c.hubClusterLister.Get(c.clusterName)
+	if err != nil {
+		return fmt.Errorf("unable to get managed cluster with name %q from hub: %w", c.clusterName, err)
+	}
+
+	// current managed cluster does not join the hub yet, do nothing.
+	if !meta.IsStatusConditionTrue(managedCluster.Status.Conditions, clusterv1.ManagedClusterConditionJoined) {
+		syncCtx.Recorder().Eventf("ManagedClusterIsNotAccepted", "Managed cluster %q does not join the hub yet", c.clusterName)
```
deads2k, line 71: has not joined the hub yet

pkg/spoke/managedcluster/claim_controller.go
```diff
@@ -0,0 +1,140 @@
...
+		return reservedClaims[i].Name < reservedClaims[j].Name
+	})
+
+	sort.SliceStable(customizedClaims, func(i, j int) bool {
+		return customizedClaims[i].Name < customizedClaims[j].Name
+	})
+
+	// merge and truncated claims
+	claims := append(reservedClaims, customizedClaims...)
+	if total := len(claims); total > c.maxClusterClaims {
```
deads2k, line 114: recommendation: express this as the max number of custom claims.  It seems like you should always be able to update reserved names

pkg/spoke/spokeagent.go
```diff
@@ -276,6 +296,8 @@ func (o *SpokeAgentOptions) AddFlags(fs *pflag.FlagSet) {
 		"A list of reachable spoke cluster api server URLs for hub cluster.")
 	fs.DurationVar(&o.ClusterHealthCheckPeriod, "cluster-healthcheck-period", o.ClusterHealthCheckPeriod,
 		"The period to check managed cluster kube-apiserver health")
+	fs.IntVar(&o.MaxClusterClaims, "cluster-claims-max", o.MaxClusterClaims,
```
deads2k, line 299: I think you probably want to have this as the max custom cluster claims.

deads2k:
I have recommendations that can come as followups

/lgtm

PR #105, opened 2020-12-17
Refine the logic to get agent name

elgnay:
The related issue: https://github.com/open-cluster-management/backlog/issues/7679

### DIFF
--- a/pkg/spoke/spokeagent.go
+++ b/pkg/spoke/spokeagent.go
@@ -16,6 +16,7 @@ import (
 	"github.com/open-cluster-management/registration/pkg/spoke/managedcluster"
 
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/library-go/pkg/operator/events"
 
 	"github.com/spf13/pflag"
 
@@ -26,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -84,7 +86,13 @@ func NewSpokeAgentOptions() *SpokeAgentOptions {
 // create a valid hub kubeconfig. Once the hub kubeconfig is valid, the
 // temporary controller is stopped and the main controllers are started.
 func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext *controllercmd.ControllerContext) error {
-	if err := o.Complete(); err != nil {
+	// create kube client
+	spokeKubeClient, err := kubernetes.NewForConfig(controllerContext.KubeConfig)
+	if err != nil {
+		return err
+	}
+
+	if err := o.Complete(spokeKubeClient.CoreV1(), ctx, controllerContext.EventRecorder); err != nil {
 		klog.Fatal(err)
 	}
 
@@ -94,11 +102,7 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
 
 	klog.Infof("Cluster name is %q and agent name is %q", o.ClusterName, o.AgentName)
 
-	// create kube client and shared informer factory for spoke cluster
-	spokeKubeClient, err := kubernetes.NewForConfig(controllerContext.KubeConfig)
-	if err != nil {
-		return err
-	}
+	// create shared informer factory for spoke cluster
 	spokeKubeInformerFactory := informers.NewSharedInformerFactory(spokeKubeClient, 10*time.Minute)
 
 	// get spoke cluster CA bundle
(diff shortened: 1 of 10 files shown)


pkg/spoke/spokeagent.go
```diff
@@ -388,9 +413,18 @@ func (o *SpokeAgentOptions) hasValidHubClientConfig() (bool, error) {
 //
 //   1. take clusterName from input arguments if it is not empty
 //   2. TODO: read cluster name from openshift struct if the spoke agent is running in an openshift cluster
-//   3. take cluster/agent names from the mounted secret if they exist
-//   4. generate random cluster/agent names then
+//   3. take cluster/agent name from tls certification if it exist
+//   4. take cluster/agent names from the mounted secret if they exist
```
qiujian16: we need a clearer doc here  that we find cluster/agent name from cert at first,  and fallback to use values in secret if it does not exists, and fall back to generate them

pkg/spoke/spokeagent.go
```diff
@@ -399,23 +433,39 @@ func (o *SpokeAgentOptions) getOrGenerateClusterAgentNames() (string, string) {
...
 		clusterNameBytes, err := ioutil.ReadFile(path.Clean(clusterNameFilePath))
-		if err != nil {
-			// generate random cluster name if faild
-			clusterName = generateClusterName()
-		} else {
+		switch {
+		case len(clusterNameInCert) > 0:
+			// use cluster name loaded from the tls certification
+			clusterName = clusterNameInCert
+			if err != nil && clusterNameInCert != string(clusterNameBytes) {
```
skeeey: this should be `err == nil`, and we may use a new var name for this error

pkg/spoke/spokeagent.go
```diff
@@ -399,23 +433,39 @@ func (o *SpokeAgentOptions) getOrGenerateClusterAgentNames() (string, string) {
...
 	var agentName string
-	if err != nil {
-		// generate random agent name if faild
-		agentName = generateAgentName()
-	} else {
+	switch {
+	case len(agentNameInCert) > 0:
+		// use agent name loaded from the tls certification
+		agentName = agentNameInCert
+		if err != nil && agentNameInCert != string(agentNameBytes) {
```
skeeey: this should be `err == nil`, and we may use a new var name for this error

pkg/spoke/spokeagent.go
elgnay: Done!

pkg/spoke/spokeagent.go
elgnay: Fixed.

pkg/spoke/spokeagent.go
elgnay: Fixed.

test/integration/spokeagent_restart_test.go
```diff
@@ -0,0 +1,310 @@
...
+		gomega.Eventually(func() bool {
+			_, err = util.FindAutoApprovedSpokeCSR(kubeClient, managedClusterName)
+			if err != nil {
+				return false
+			}
+			return true
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeTrue())
+	})
+
+	ginkgo.It("restart agent with a different cluster name", func() {
```
qiujian16, line 166: this is to test that cluster name in  secret is changed? I think it needs a better description

test/integration/spokeagent_restart_test.go
```diff
@@ -0,0 +1,310 @@
...
+			}
+			return true
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeTrue())
+
+		ginkgo.By("Stop registration agent and wait for a grace period")
+		stopAgent()
+		time.Sleep(5 * time.Second)
+
+		// remove the join condition
+		spokeCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
```
qiujian16, line 105: why do we need to remove the join condition?

pkg/spoke/spokeagent.go
```diff
@@ -399,23 +440,39 @@ func (o *SpokeAgentOptions) getOrGenerateClusterAgentNames() (string, string) {
...
 	var agentName string
-	if err != nil {
-		// generate random agent name if faild
-		agentName = generateAgentName()
-	} else {
+	switch {
+	case len(agentNameInCert) > 0:
+		// use agent name loaded from the tls certification
+		agentName = agentNameInCert
+		if err == nil && agentNameInCert != string(agentNameBytes) {
```
qiujian16: i am not sure we need err==nil here, I think we output the warning whenever it is different

test/integration/spokeagent_restart_test.go
elgnay: User is allow to specify a cluster name when start registration agent with flag `cluster-name`. If the agent has already had a hub kubecofig with different cluster name, this will trigger a bootstrap process. That 's the case I want to test with this test case.

test/integration/spokeagent_restart_test.go
elgnay: If a cluster has a join condition we think the registration agent works well as expected. Before restarting, it has already had a  join condition. That's the reason why I have to remove this condition.

pkg/spoke/spokeagent.go
elgnay: It makes sense. Whenever the cluster/agent name in cert is used, we should output a warning message.

test/integration/spokeagent_restart_test.go
qiujian16: Ok, I think we at least need some doc to describe the case.

test/integration/spokeagent_restart_test.go
qiujian16: i think it will be better to add some doc here

test/integration/spokeagent_restart_test.go
elgnay: Doc added.

test/integration/spokeagent_restart_test.go
elgnay: Doc added.

PR #109, opened 2021-01-15
List-watches secret via namespaced informer to tighten rbac scope

yue9944882:
before this pull, spoke agent will do cluster-scoped list-watching agaisnt secret resource which requires cluster-scoped RBAC permission. this can infeasible in a larger cluster shared by multiple tenants. this pull tightens rbac permission by list-watching secrets w/i the target namespace..

### DIFF
--- a/pkg/spoke/spokeagent.go
+++ b/pkg/spoke/spokeagent.go
@@ -104,6 +104,7 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
 
 	// create shared informer factory for spoke cluster
 	spokeKubeInformerFactory := informers.NewSharedInformerFactory(spokeKubeClient, 10*time.Minute)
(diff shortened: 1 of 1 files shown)


qiujian16 (commented):
/approve
/lgtm

pkg/spoke/spokeagent.go
```diff
@@ -172,6 +173,7 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
 
 		go bootstrapInformerFactory.Start(bootstrapCtx.Done())
 		go spokeKubeInformerFactory.Start(bootstrapCtx.Done())
+		go namespacedSpokeKubeInformerFactory.Start(bootstrapCtx.Done())
```
qiujian16, line 175: i think we need to restart the  namespacedSpokeKubeInformerFactory at line-278,  And we do not need to start spokeKubeInformerFactory during bootstrap. The spokeKubeInformerFactory will be started after bootstrap to retrieve nodes. So line-175 can also be removed.

yue9944882: thanks for the clue, updated

qiujian16 (commented):
/lgtm

PR #110, opened 2021-01-21
Use Lister to reduce the number of calls to api server

elgnay:
The related issue: https://github.com/open-cluster-management/backlog/issues/8687

### DIFF
--- a/pkg/spoke/managedcluster/creating_controller.go
+++ b/pkg/spoke/managedcluster/creating_controller.go
@@ -11,9 +11,9 @@ import (
 
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
-
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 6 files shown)


pkg/spoke/managedcluster/creating_controller.go
```diff
@@ -43,9 +44,13 @@ func NewManagedClusterCreatingController(
 		spokeCABundle:           spokeCABundle,
 		hubClusterClient:        hubClusterClient,
 	}
+
+	// create a random interval between CreatingControllerSyncInterval and CreatingControllerSyncInterval * 2.
+	interval := CreatingControllerSyncInterval + time.Duration(rand.Float64()*float64(CreatingControllerSyncInterval))
```
qiujian16: could use wait.Jitter I think

elgnay: Updated.

qiujian16 (commented):
/approve
/lgtm

PR #112, opened 2021-01-26
Defining a common user-group to match all spoke-clusters

yue9944882:
so that we can grant extra common RBAC permissions to the CSR-generated certificates mounted by the spoke-agent

@qiujian16

### DIFF
--- a/pkg/hub/csr/controller.go
+++ b/pkg/hub/csr/controller.go
@@ -160,8 +160,8 @@ func isSpokeClusterClientCertRenewal(csr *certificatesv1beta1.CertificateSigning
 	}
 
 	requestingOrgs := sets.NewString(x509cr.Subject.Organization...)
-	if requestingOrgs.Has(user.SpokeClustersGroup) { // optional common group for backward-compatibility
-		requestingOrgs.Delete(user.SpokeClustersGroup)
+	if requestingOrgs.Has(user.ManagedClustersGroup) { // optional common group for backward-compatibility
+		requestingOrgs.Delete(user.ManagedClustersGroup)
 	}
 	if requestingOrgs.Len() != 1 {
 		return false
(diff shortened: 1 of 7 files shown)


pkg/spoke/hubclientcert/controller.go
```diff
@@ -39,6 +39,7 @@ const (
 	TLSCertFile = "tls.crt"
 
 	subjectPrefix         = "system:open-cluster-management:"
+	groupAllSpokeClusters = subjectPrefix + "spoke-clusters"
```
qiujian16: maybe spokeClustersGroup with a comment that all spoke is in this group?

qiujian16 (commented):
/ok-to-test

qiujian16:
/approve

@skeeey would you also take a look? It seems secure to me

pkg/hub/user/identity.go
```diff
@@ -0,0 +1,8 @@
+package user
+
+const (
+	// SubjectPrefix is a prefix for marking open-cluster-management users
+	SubjectPrefix = "system:open-cluster-management:"
+	// SpokeClustersGroup is a common group for all spoke clusters
+	SpokeClustersGroup = SubjectPrefix + "spoke-clusters"
```
elgnay: minor: Use "managedclusters" or "managed-clusters" instead of "spoke-clusters"

pkg/hub/csr/controller.go
```diff
@@ -158,16 +159,20 @@ func isSpokeClusterClientCertRenewal(csr *certificatesv1beta1.CertificateSigning
 		return false
 	}
 
-	if len(x509cr.Subject.Organization) != 1 {
+	requestingOrgs := sets.NewString(x509cr.Subject.Organization...)
+	if requestingOrgs.Has(user.SpokeClustersGroup) { // optional common group for backward-compatibility
```
skeeey: would you add some ut for this?

yue9944882: test cases added covering the latest and the previous format of CSR

pkg/hub/user/identity.go
yue9944882: done, dont have a strong preference between the names tho

PR #115, opened 2021-02-24
Introducing ClusterClaim feature gate to opt-out from list-watching cluster-scoped resources

### LINKED ISSUE #114: Opt-out from cluster-claim controller for light-weighting managed-cluster installation
in some cases, we don't have sufficient permission to install cluster-scoped resources such as CRDs (i.e. cluster-claim) into the managed-cluster. for now, i can basically think of 2 approaches to opt-out from cluster-claim controller from running:

- 1. a simple approach will be adding a flag `--disable-cluster-claim-controller` which stops the spoke-agent from list-watching cluster-claim. 
- 2. a more generic approach is adding a flag `--controllers` which is a list of names of the controllers we expect to run. (i had a initial locally, and it will be working like the following). but the flag will only apply to spoke-agent, i wonder if we want to move further to do the similar change to hub controllers. and additionally the dark side is this approach is that AFAICT there're some mandatory controllers not expected to be disabled by any chance:

```
      --controllers stringArray                  The names of controllers to run (default [ManagedClusterController,ManagedClusterHealthCheckController,ManagedClusterLeaseController,BootstrapClientCertForHubController,ClientCertForHubController,ClusterClaimController,HubKubeconfigSecretController,ManagedClusterCreatingController])
```

@qiujian16 for thoughts?

yue9944882:
/cc @qiujian16 

Fixes: #114

### DIFF
--- a/pkg/features/feature.go
+++ b/pkg/features/feature.go
@@ -1,0 +1,44 @@
+package features
+
+import (
+	"k8s.io/component-base/featuregate"
+	"k8s.io/klog/v2"
+)
+
+const (
+	// Every feature gate should add method here following this template:
+	//
+	// // owner: @username
+	// // alpha: v1.X
+	// MyFeature featuregate.Feature = "MyFeature"
+
+	// ClusterClaim will start a new controller in the spoke-agent to manage the cluster-claim
+	// resources in the managed cluster.
+	//
+	// The cluster-claim controller is majorly for collecting claims and updating claims field
+	// in managedcluster status. When it exceeds the limit specified by "--max-custom-cluster-claims",
+	// the extra claims will be truncated.
+	//
+	// If it is disabled, the user will see empty claims field in managedcluster status. The
+	// deployer who disable the feature may need to update claim field in managed cluster status
+	// itself to avoid impact to users.
+	ClusterClaim featuregate.Feature = "ClusterClaim"
+)
+
+var (
+	// DefaultMutableFeatureGate is made up of multiple mutable feature-gates.
+	DefaultMutableFeatureGate featuregate.MutableFeatureGate = featuregate.NewFeatureGate()
+)
+
+func init() {
+	if err := DefaultMutableFeatureGate.Add(defaultRegistrationFeatureGates); err != nil {
+		klog.Fatalf("Unexpected error: %v", err)
+	}
+}
+
+// defaultRegistrationFeatureGates consists of all known ocm-registration
+// feature keys.  To add a new feature, define a key for it above and
+// add it here.
+var defaultRegistrationFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	ClusterClaim: {Default: true, PreRelease: featuregate.Beta},
+}
--- a/pkg/spoke/spokeagent.go
+++ b/pkg/spoke/spokeagent.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/openshift/library-go/pkg/controller/factory"
 	"io/ioutil"
 	"os"
 	"path"
@@ -11,6 +12,7 @@ import (
 
 	clusterv1client "github.com/open-cluster-management/api/client/cluster/clientset/versioned"
 	clusterv1informers "github.com/open-cluster-management/api/client/cluster/informers/externalversions"
+	"github.com/open-cluster-management/registration/pkg/features"
 	"github.com/open-cluster-management/registration/pkg/helpers"
 	"github.com/open-cluster-management/registration/pkg/spoke/hubclientcert"
 	"github.com/open-cluster-management/registration/pkg/spoke/managedcluster"
@@ -255,22 +257,24 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
 		o.ClusterHealthCheckPeriod,
 		controllerContext.EventRecorder,
 	)
-
 	spokeClusterClient, err := clusterv1client.NewForConfig(controllerContext.KubeConfig)
 	if err != nil {
 		return err
 	}
 	spokeClusterInformers := clusterv1informers.NewSharedInformerFactory(spokeClusterClient, 10*time.Minute)
 
-	// create managedClusterClaimController to sync cluster claims
-	managedClusterClaimController := managedcluster.NewManagedClusterClaimController(
-		o.ClusterName,
-		o.MaxCustomClusterClaims,
-		hubClusterClient,
-		hubClusterInformerFactory.Cluster().V1().ManagedClusters(),
-		spokeClusterInformers.Cluster().V1alpha1().ClusterClaims(),
-		controllerContext.EventRecorder,
-	)
+	var managedClusterClaimController factory.Controller
+	if features.DefaultMutableFeatureGate.Enabled(features.ClusterClaim) {
+		// create managedClusterClaimController to sync cluster claims
+		managedClusterClaimController = managedcluster.NewManagedClusterClaimController(
+			o.ClusterName,
+			o.MaxCustomClusterClaims,
+			hubClusterClient,
+			hubClusterInformerFactory.Cluster().V1().ManagedClusters(),
+			spokeClusterInformers.Cluster().V1alpha1().ClusterClaims(),
+			controllerContext.EventRecorder,
+		)
+	}
 
 	go hubKubeInformerFactory.Start(ctx.Done())
 	go hubClusterInformerFactory.Start(ctx.Done())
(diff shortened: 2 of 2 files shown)


pkg/features/feature.go
```diff
@@ -0,0 +1,37 @@
...
+	if err := DefaultMutableFeatureGate.Add(defaultRegistrationFeatureGates); err != nil {
+		klog.Fatalf("Unexpected error: %v", err)
+	}
+}
+
+// defaultRegistrationFeatureGates consists of all known ocm-registration
+// feature keys.  To add a new feature, define a key for it above and
+// add it here.
+var defaultRegistrationFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	ClusterClaim: {Default: true, PreRelease: featuregate.Beta},
```
yue9944882, line 43:
```
 --feature-gates mapStringBool              A set of key=value pairs that describe feature gates for alpha/experimental features. Options are:
                                                 AllAlpha=true|false (ALPHA - default=false)
                                                 AllBeta=true|false (BETA - default=false)
                                                 ClusterClaim=true|false (BETA - default=true)
```

we have to make this beta otherwise it wont be configured via command-lines @qiujian16

pkg/spoke/spokeagent.go
```diff
@@ -282,14 +273,26 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
...
 	go managedClusterHealthCheckController.Run(ctx, 1)
-	go managedClusterClaimController.Run(ctx, 1)
+	if features.DefaultMutableFeatureGate.Enabled(features.ClusterClaim) {
+		// create managedClusterClaimController to sync cluster claims
+		managedClusterClaimController := managedcluster.NewManagedClusterClaimController(
+			o.ClusterName,
+			o.MaxCustomClusterClaims,
+			hubClusterClient,
+			hubClusterInformerFactory.Cluster().V1().ManagedClusters(),
+			spokeClusterInformers.Cluster().V1alpha1().ClusterClaims(),
```
qiujian16: declaring  informers from factory after factory is started might be a problem.

qiujian16:
@deads2k Would you also  review it?

pkg/features/feature.go
qiujian16: it is ok to make this as beta.

pkg/spoke/spokeagent.go
yue9944882: good catch, it seems that cluster-claim controller is the only one that actually references `spokeClusterInformers`, so 
i wrapped that into the feature-gated block.

pkg/features/feature.go
```diff
@@ -0,0 +1,37 @@
...
+
+const (
+	// Every feature gate should add method here following this template:
+	//
+	// // owner: @username
+	// // alpha: v1.X
+	// MyFeature featuregate.Feature = "MyFeature"
+
+	// ClusterClaim will manage the cluster-claim resources in the managed cluster. Disabling the
+	// the feature-gate can make the spoke-agent run without installing cluster-claim CRDs in the
```
elgnay: The cluster claim CRD is installed even when the feature gate is disabled unless we put the feature gate into Klusterlet CR and let registration-operator decide if cluster claim CRD should be installed or not.

yue9944882: for clarification, the point of the feature-gate is not for chopping cluster-claim off intentionally from a managed-cluster. it's to make the agent work when the CRD cannot be installed due to insufficient permission. in our case, the crux is that neither registration-operator nor OCM admin doesn't have the cluster-scoped access to the managed cluster --- it doesn't actually have a choice between installation CRDs or not.

pkg/spoke/spokeagent.go
```diff
@@ -260,36 +261,38 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
 	if err != nil {
 		return err
 	}
-	spokeClusterInformers := clusterv1informers.NewSharedInformerFactory(spokeClusterClient, 10*time.Minute)
```
deads2k, line 263: I would keep the informers here and move your usage of them into the feature gate block.

pkg/spoke/spokeagent.go
```diff
@@ -260,36 +261,38 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
...
-		hubClusterInformerFactory.Cluster().V1().ManagedClusters(),
-		spokeClusterInformers.Cluster().V1alpha1().ClusterClaims(),
-		controllerContext.EventRecorder,
-	)
 
 	go hubKubeInformerFactory.Start(ctx.Done())
 	go hubClusterInformerFactory.Start(ctx.Done())
 	go spokeKubeInformerFactory.Start(ctx.Done())
 	go namespacedSpokeKubeInformerFactory.Start(ctx.Done())
-	go spokeClusterInformers.Start(ctx.Done())
```
deads2k: I would keep this with the others.  An unused informer .start is a no-op.

pkg/features/feature.go
deads2k: My memory of the controller is that it creates `ManagedClusterClaim`, not a CRD.  Is this about the ManagedClusterClaim CR or the ManagedClusterClaim CRD?

deads2k: > My memory of the controller is that it creates `ManagedClusterClaim`, not a CRD. Is this about the ManagedClusterClaim CR or the ManagedClusterClaim CRD?

Related: can we update the comment to indicate
1. the end-user-impact of this being disabled
2. how an end-user knows this is disabled
3. how a deployer can mitigate this impact

pkg/spoke/spokeagent.go
yue9944882: done

pkg/spoke/spokeagent.go
yue9944882: updated

pkg/features/feature.go
yue9944882: > Is this about the ManagedClusterClaim CR or the ManagedClusterClaim CRD?

i believe @elgnay is talking about the CRD. 

> My memory of the controller is that it creates ManagedClusterClaim, not a CRD. 

as far as i can see from the code, the cluster-claim controller doesnt actually create the cluster-claim, it just bounds the # of custom cluster-claims in the managed cluster by setting `--max-custom-cluster-claims` of the agent --- i'm a bit confused at which component is creating the claim? the intention of ManagedClusterClaim CRD is even unclear to me either. 🤔

> Related: can we update the comment to indicate

just enrich the doc for the feature-gate, PTAL

elgnay: @yue9944882 You can find the intention of ManagedClusterClaim from this [enhancement proposal](https://github.com/open-cluster-management/enhancements/tree/master/enhancements/sig-architecture/4-cluster-claims). Registration agent focus on vendor-neutral logic. It only collects all claims on managed cluster and puts them into status of managed cluster on hub. The logic of cluster claim generation might be vendor/product specific. It now been done by `work-manager`, an addon running on managed cluster. Of course, claims could be created by any other agent as well. So as a summary, 
- ManagedClusterClaim CRD is installed by klusterlet (https://github.com/open-cluster-management/registration-operator);
- and ManagedClusterClaim CRs are created by work-manager addon (https://github.com/open-cluster-management/multicloud-operators-foundation);
- Registration agent watchs ManagedClusterClaim CRs and sync them to ManagedCluster on hub;

pkg/features/feature.go
```diff
@@ -0,0 +1,42 @@
...
+	// Every feature gate should add method here following this template:
+	//
+	// // owner: @username
+	// // alpha: v1.X
+	// MyFeature featuregate.Feature = "MyFeature"
+
+	// ClusterClaim will start a new controller in the spoke-agent to manage the cluster-claim
+	// resources in the managed cluster.
+	//
+	// The cluster-claim controller is majorly for truncating the existing custom cluster-claims
```
qiujian16: it is for collecting claims and updating claims field in managedcluster status. If it is disabled, the user will see empty claims field in managedcluster status. The deployer who disable the feature may need to update claim field in managed cluster status itself to avoid impact to users.

pkg/features/feature.go
yue9944882: > ManagedClusterClaim CRD is installed by klusterlet (https://github.com/open-cluster-management/registration-operator);

isnt registration-operator an completely optional component? e.g. prometheus-operator is optional for automating deployment for prometheus. the automation for installing ManagedClusterClaim CRD shouldnt be an blocking issue for introducing the feature-gate.

pkg/features/feature.go
yue9944882: applied

pkg/features/feature.go
elgnay: > isnt registration-operator an completely optional component? 

Yes, it's optional for running registration agent. You are able to deploy it in your own way.

qiujian16:
/lgtm

Thanks!

PR #116, opened 2021-03-08
Using clusterrole instead of current registration roles

skeeey:


skeeey:
refer to https://github.com/open-cluster-management/backlog/issues/9057

pkg/hub/clusterrole/controller.go
```diff
@@ -0,0 +1,92 @@
...
+	clusterInformer clusterv1informer.ManagedClusterInformer,
+	clusterRoleInformer rbacv1informers.ClusterRoleInformer,
+	recorder events.Recorder) factory.Controller {
+	c := &clusterroleController{
+		kubeClient:    kubeClient,
+		clusterLister: clusterInformer.Lister(),
+		eventRecorder: recorder.WithComponentSuffix("managed-cluster-clusterrole-controller"),
+	}
+	return factory.New().
+		WithInformers(clusterInformer.Informer(), clusterRoleInformer.Informer()).
```
qiujian16: adding filter here will be better.
I think you also need clusterinformer?

When it is triggered when we create a new cluster?
When it is triggered when there are no clusters.

skeeey: OK, I will adding a filer here and I already added the `clusterinformer`

pkg/hub/clusterrole/controller.go
```diff
@@ -0,0 +1,107 @@
...
+			func(name string) ([]byte, error) {
+				return assets.MustCreateAssetFromTemplate(name, bindata.MustAsset(filepath.Join(manifestDir, name)), nil).Data, nil
+			},
+			clusterRoleFiles...,
+		)
+	}
+
+	// Make sure the managedcluser cluserroles are existed if there are clusters
+	results := resourceapply.ApplyDirectly(
+		resourceapply.NewKubeClientHolder(c.kubeClient),
```
qiujian16, line 91: should we consider getting clusterrole/binding from lister at first?

skeeey: yeah, I think so, I will update this

pkg/hub/clusterrole/controller.go
```diff
@@ -0,0 +1,135 @@
...
+			clusterRoleFiles...,
+		)
+	}
+
+	// Make sure the managedcluser cluserroles are existed if there are clusters
+	appliedClusterRoleFiles := []string{}
+	_, err = c.clusterRoleLister.Get(registrationClusterRole)
+	switch {
+	case errors.IsNotFound(err):
+		appliedClusterRoleFiles = append(appliedClusterRoleFiles, clusterRoleFiles[0])
```
qiujian16: do not use index here, you can have two consts for each file.

pkg/hub/clusterrole/controller.go
```diff
@@ -0,0 +1,135 @@
...
+	case err != nil:
+		return err
+	}
+
+	_, err = c.clusterRoleLister.Get(workClusterRole)
+	switch {
+	case errors.IsNotFound(err):
+		appliedClusterRoleFiles = append(appliedClusterRoleFiles, clusterRoleFiles[1])
+	case err != nil:
+		return err
```
qiujian16: should we also compare the rules in clusterroles?

pkg/hub/clusterrole/controller.go
qiujian16: i will take it back...we did not able to compare the roles with lister. I think we should start with Get at first.

qiujian16 (commented):
/approve
/lgtm

PR #118, opened 2021-03-25
using livez instead of readyz

skeeey:


skeeey:
refer to https://github.com/open-cluster-management/backlog/issues/10823

PR #120, opened 2021-03-29
Refactor and make client cert controller reusable

elgnay:
As @qiujian16 suggested, it is split into two PRs to support addon registration. This is the first one. 

The related issue: https://github.com/open-cluster-management/backlog/issues/10714


pkg/clientcert/cert_controller.go
```diff
@@ -0,0 +1,329 @@
...
+var ControllerResyncInterval = 5 * time.Minute
+
+// CertificateManager abstracts behaviours which might be different during the creation/rotation of client
+// certification.
+type CertificateManager interface {
+	// Owner returns the name of the client certificate owner, which is used by clientCertificateController
+	// in messages of log/event.
+	Owner() string
+
+	// NewCSR returns an new object of csr. And then clientCertificateController will create it on api server.
```
qiujian16: hub apiserver

pkg/clientcert/cert_controller.go
```diff
@@ -0,0 +1,329 @@
...
+	ClusterNameLabel = "open-cluster-management.io/cluster-name"
+	AddonNameLabel   = "open-cluster-management.io/addon-name"
+	SignerNameLabel  = "open-cluster-management.io/signer-name"
+)
+
+// ControllerResyncInterval is exposed so that integration tests can crank up the constroller sync speed.
+var ControllerResyncInterval = 5 * time.Minute
+
+// CertificateManager abstracts behaviours which might be different during the creation/rotation of client
+// certification.
```
qiujian16: I think you may add some doc about cluster registration and addon registration will implement the  interface.

pkg/clientcert/certificate.go
```diff
@@ -1,35 +1,34 @@
...
+//   1. KubeconfigFile exists when hasKubeconfig is true
 //   2. TLSKeyFile exists
 //   3. TLSCertFile exists and the certificate is not expired
-//   4. If not empty, the given commonName matches the common name of the subject in the
-//      certificate stored in TLSCertFile
-func hasValidKubeconfig(secret *corev1.Secret, commonName string) bool {
-	if secret.Data == nil {
-		klog.V(4).Infof("No kubeconfig found in secret %q", secret.Namespace+"/"+secret.Name)
+//   4. If subject is specified, it matches the subject in the certificate stored in TLSCertFile
+func HasValidClientCertificate(secret *corev1.Secret, hasKubeconfig bool, subject *pkix.Name) bool {
```
qiujian16: why don't we have another hasValidKubeConfig func?

pkg/features/feature.go
```diff
@@ -23,6 +23,8 @@ const (
 	// deployer who disable the feature may need to update claim field in managed cluster status
 	// itself to avoid impact to users.
 	ClusterClaim featuregate.Feature = "ClusterClaim"
+
```
qiujian16: add some doc here, what user will see when it is enabled.

pkg/spoke/addon/certificate_manager.go
```diff
@@ -0,0 +1,251 @@
...
+func (m *addonCertificateManager) GetCSRClient() (csrclient.CertificateSigningRequestInterface, error) {
+	// if addon has no bootstrap secret, always return the default csr client (created with registration
+	// agent's hub kubeconfig)
+	if len(m.config.BootstrapSecretName) == 0 {
+		return m.hubCSRClient, nil
+	}
+
+	// if the addon has already had a valid client cert (bootstraped), return the default
+	// csr client for cert renewal
+	secret, err := m.kubeClient.CoreV1().Secrets(m.config.InstallationNamespace).Get(context.Background(), m.config.SecretName, metav1.GetOptions{})
```
qiujian16: I think we alway use the registration csr client? How can we ensure that the addon's client cert can be used for csr.

pkg/clientcert/cert_controller.go
```diff
@@ -0,0 +1,329 @@
...
+	NewCSR(keyData []byte) (*certificates.CertificateSigningRequest, error)
+
+	// AddAdditionalSecretData appends addintional data into generated client certificate secret other than
+	// tls.key/tls.crt, which are added by clientCertificateController. The addintional data may include
+	// kubeconfig and cluster/agent name.
+	AddAdditionalSecretData(data map[string][]byte) error
+
+	// HasValidClientCertificate checks the given secret and returns true if it contains a valid client
+	// certificate.
+	HasValidClientCertificate(secret *corev1.Secret) bool
```
qiujian16: I would call it HasValidRegistrationSecret

pkg/spoke/addon/certificate_manager.go
```diff
@@ -0,0 +1,251 @@
...
+				certificates.UsageClientAuth,
+			},
+			SignerName: &m.config.SignerName,
+		},
+	}
+
+	return csr, nil
+}
+
+func (m *addonCertificateManager) HasValidClientCertificate(secret *corev1.Secret) bool {
```
qiujian16: i would compare kubeconfig and clientcert separately.

pkg/spoke/addon/configuration.go
```diff
@@ -0,0 +1,222 @@
...
+	installationNamespaceAnnotation = "installNamespace"
+	bootstrapSecretAnnotation       = "bootstrapSecret"
+	registrationAnnotation          = "registrations"
+	clusterNamePlaceHolder          = "{ClusterName}"
+	agentNamePlaceHolder            = "{AgentName}"
+)
+
+// registrationConfig contains necessary information for addon registration. A addon may have multiple
+// registrationConfigs.
+type registrationConfig struct {
```
qiujian16: You may want this to be a public class that shared with leases part. FYI @skeeey

pkg/spoke/addon/configuration.go
```diff
@@ -0,0 +1,222 @@
...
+	AddonName             string      `json:"addonName"`
+	InstallationNamespace string      `json:"installNamespace"`
+	BootstrapSecretName   string      `json:"bootstrapSecretName,omitempty"`
+	SignerName            string      `json:"signerName,omitempty"`
```
qiujian16: i think we will support multiple signers?

It is better to support that in API at the very beginning, and add a todo to implement later.

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,203 @@
...
+			hubAddonInformers.Informer()).
+		WithSync(c.sync).
+		ToController("AddonClientCertManagerController", recorder)
+}
+
+func (c *addonRegistrationController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	addonName := syncCtx.QueueKey()
+	klog.V(4).Infof("Reconciling addon %q", addonName)
+
+	cachedConfigs := c.addonRegistrationConfigs[addonName]
```
qiujian16: do we need to keep the cached configs for cleanup? We may not need to handle that because cert will expire and hub controller to setup rbac to disallow the cert. 

The  problem here is if this is restarted, we still cannot guarantee the cleanup.

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,203 @@
...
+		hubClientConfig:      hubClientConfig,
+		hubKubeClient:        hubKubeClient,
+		hubAddonLister:       hubAddonInformers.Lister(),
+
+		hubCSRInformer: hubCSRInformer,
+		hubCSRClient:   hubCSRClient,
+
+		recorder: recorder,
+
+		addonRegistrationConfigs: map[string][]registrationConfig{},
```
qiujian16: let's add some doc on what is the key of the two maps.

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,203 @@
...
+		addonCertificateManagers: map[string]*addonCertificateManager{},
```
qiujian16: I think you can use registrationConfig as the key directly instead of hash.

pkg/spoke/addon/configuration.go
skeeey: public is unnecessary, we should be in the same package

skeeey: could we consider this to
```
type registration struct {
    SignerName            string      `json:"signerName,omitempty"`
    Subject               certSubject `json:"subject,omitempty"`

}

type registrationConfig struct {
     AddonName             string      `json:"addonName"`
	InstallationNamespace string      `json:"installNamespace"`
	BootstrapSecretName   string      `json:"bootstrapSecretName,omitempty"`
        Registrations []registration
        ...
}
```

thus `getRegistrationConfigs` will return one registrationConfig for one addon

qiujian16:
based on our discussion, let's split this to two PRs.
1. refactor code to build a common controller for csr
2. implement addon manager for csr

pkg/clientcert/cert_controller.go
```diff
@@ -0,0 +1,317 @@
...
+
+func NewClientCertificateController(
+	clientCertSecretNamespace string,
+	clientCertSecretName string,
+	hubCSRInformer certificatesinformers.CertificateSigningRequestInformer,
+	hubCSRClient csrclient.CertificateSigningRequestInterface,
+	spokeSecretInformer corev1informers.SecretInformer,
+	spokeCoreClient corev1client.CoreV1Interface,
+	csrSubject *pkix.Name,
+	newCSRFunc NewCSRFunc,
```
qiujian16: If the only reason that we have this func is because of signer, i would directly use signer as param here.

pkg/clientcert/cert_controller.go
```diff
@@ -0,0 +1,317 @@
...
+	extendClientCertSecretDataFunc ExtendClientCertSecretDataFunc,
```
qiujian16: how about having
```
additonalSecretData map[string][]byte
```
?

pkg/clientcert/cert_controller.go
elgnay: When creating a csr, the following parts can be customized:
- name/generated name;
- labels
- subject (cn, o, ou...)
- signer

Besides subject, we still need 3 parameters to cover all of them. So passing a function might be better in this situation.

pkg/clientcert/cert_controller.go
elgnay: A static map might work as well. With the original logic, some data will be added when a certian condition is met. e.g. kubeconfig is added once tls.key/tls.crt are available in the secret. Anyway, I think it's fine to always add all data from a static map in the secret.

pkg/clientcert/cert_controller.go
elgnay: I replaced the func with 3 parameters below:
```
csrObjectMeta metav1.ObjectMeta
csrSubject    *pkix.Name
csrSignerName string
```

pkg/clientcert/cert_controller.go
elgnay: Replaced the func with a static map.

pkg/clientcert/cert_controller.go
```diff
@@ -0,0 +1,344 @@
...
+
+	certificates "k8s.io/api/certificates/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	certificatesinformers "k8s.io/client-go/informers/certificates/v1beta1"
+	corev1informers "k8s.io/client-go/informers/core/v1"
+	csrclient "k8s.io/client-go/kubernetes/typed/certificates/v1beta1"
```
deads2k, line 23: yikes.  This will need updating to v1 in another PR

pkg/clientcert/cert_controller.go
```diff
@@ -0,0 +1,344 @@
...
+			return accessor.GetName()
+		}, csrFilterFunc, hubCSRInformer.Informer()).
+		WithSync(c.sync).
+		ResyncEvery(ControllerResyncInterval).
+		ToController(controllerName, recorder)
+}
+
+func (c *clientCertificateController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	// get secret containing client certificate
+	secret, err := c.spokeCoreClient.Secrets(c.clientCertSecretNamespace).Get(ctx, c.clientCertSecretName, metav1.GetOptions{})
```
deads2k: followup: as you start running more of these, you may end up wanting to use listers.

deads2k:
This is a very good start.  I think before calling it finished, you'll want to find a way to make `NewClientCertificateController` less intimidating to a newcomer.  I would probably start by documenting each arg.  And then consider trying to bundle some related bits into structs you can check for legal values: one for CSRs maybe, one for clients, one for targets.

pkg/clientcert/cert_controller.go
elgnay: Yes, the API upgrade will be done in another PR.

pkg/clientcert/cert_controller.go
```diff
@@ -0,0 +1,350 @@
...
+
+// ClientCertOption includes options that is used to create client certificate
+type ClientCertOption struct {
+	// SecretNamespace is the namespace of the secret containing client certificate.
+	SecretNamespace string
+	// SecretName is the name of the secret containing client certificate. The secret will be created if
+	// it does not exist.
+	SecretName string
+	// AdditonalSecretData contains data that will be added into client certificate secret besides tls.key/tls.crt
+	AdditonalSecretData map[string][]byte
```
qiujian16, line 74: i think csrFilterFunc should also be put here?

elgnay: Updated as suggested.

PR #123, opened 2021-04-01
Support addon registration

elgnay:
The related issue: https://github.com/open-cluster-management/backlog/issues/10713

### DIFF
--- a/pkg/spoke/addon/configuration.go
+++ b/pkg/spoke/addon/configuration.go
@@ -1,18 +1,74 @@
 package addon
 
 import (
+	"context"
+	"crypto/sha256"
+	"crypto/x509/pkix"
+	"encoding/json"
+	"fmt"
+	"strings"
+
 	addonv1alpha1 "github.com/open-cluster-management/api/addon/v1alpha1"
+	certificatesv1beta1 "k8s.io/api/certificates/v1beta1"
 )
 
 const (
 	defaultAddOnInstallationNamespace = "open-cluster-management-agent-addon"
 	installationNamespaceAnnotation   = "addon.open-cluster-management.io/installNamespace"
+	// registrationAnnotation includes the addOn registration configuration. If specified, it means the registration is enabled.
+	// The value is a json string, like
+	// `[ { "signerName":"signer1", "subject":{ "commonName": "addon1-agent", "organization":["addon1"], "organizationalUnit":["acm"] }]`
+	// where subject is optional.
+	registrationAnnotation = "addon.open-cluster-management.io/registrations"
 )
 
 // addonConfig contains addon configuration information.
 type addOnConfig struct {
 	AddOnName             string
 	InstallationNamespace string
+}
+
+// registrationConfig contains necessary information for addon registration
+// TODO: Refactor the code here once the registration configuration is available in spec of ManagedClusterAddOn
+type registrationConfig struct {
+	AddOnName             string `json:"addOnName,omitempty"`
+	InstallationNamespace string `json:"installNamespace,omitempty"`
+	// SignerName is used to create csr. The default value is "kubernetes.io/kube-apiserver-client".
+	SignerName string      `json:"signerName,omitempty"`
+	Subject    certSubject `json:"subject,omitempty"`
+
+	// secretName is the name of secret containing client certificate. If the SignerName is "kubernetes.io/kube-apiserver-client",
+	// the secret name will be "{addon name}-hub-kubeconfig". Otherwise, the secret name will be "{addon name}-{signer name}-client-cert".
+	secretName string
+	hash       string
+	stopFunc   context.CancelFunc
+}
+
+func (c *registrationConfig) x509Subject(clusterName, agentName string) *pkix.Name {
+	subject := &pkix.Name{
+		CommonName:         c.Subject.CommonName,
+		Organization:       c.Subject.Organization,
+		OrganizationalUnit: c.Subject.OrganizationalUnit,
+	}
+
+	// set the default common name
+	if len(subject.CommonName) == 0 {
+		subject.CommonName = defaultCommonName(clusterName, agentName, c.AddOnName)
+	}
+
+	// set the default organization if signer is KubeAPIServerClientSignerName
+	if c.SignerName == certificatesv1beta1.KubeAPIServerClientSignerName && len(subject.Organization) == 0 {
+		subject.Organization = []string{defaultOrganization(clusterName, c.AddOnName)}
+	}
+
+	return subject
+}
+
+// certSubject contains fields which are used to generate csr subject and can be customized by users
+type certSubject struct {
+	CommonName         string   `json:"commonName,omitempty"`
+	Organization       []string `json:"organization,omitempty"`
+	OrganizationalUnit []string `json:"organizationalUnit,omitempty"`
 }
 
 // getAddOnConfig returns addon configuration information from addon annotations.
@@ -27,7 +83,63 @@ func getAddOnConfig(addOn *addonv1alpha1.ManagedClusterAddOn) (*addOnConfig, err
 		InstallationNamespace: installationNamespace,
 	}
 
-	//TODO fill out the bootstrap secret name and registrions information
-
 	return addOnConfig, nil
+}
+
+// getRegistrationConfigs reads annotations of a addon and returns a map of registrationConfig whose
+// key is the hash of the registrationConfig
+func getRegistrationConfigs(addOnName string, annotations map[string]string) (map[string]registrationConfig, error) {
+	ns := annotations[installationNamespaceAnnotation]
+	if len(ns) == 0 {
+		return nil, nil
+	}
+
+	registration := annotations[registrationAnnotation]
+	if len(registration) == 0 {
+		return nil, nil
+	}
+
+	// parse registration configs from annotation
+	items := []registrationConfig{}
+	err := json.Unmarshal([]byte(registration), &items)
+	if err != nil {
+		return nil, err
+	}
+
+	configs := map[string]registrationConfig{}
+	for _, item := range items {
+		item.AddOnName = addOnName
+		item.InstallationNamespace = ns
+
+		// set the default signer name
+		if len(item.SignerName) == 0 {
+			item.SignerName = certificatesv1beta1.KubeAPIServerClientSignerName
+		}
+		// set the secret name of client certificate
+		switch item.SignerName {
+		case certificatesv1beta1.KubeAPIServerClientSignerName:
+			item.secretName = fmt.Sprintf("%s-hub-kubeconfig", addOnName)
+		default:
+			item.secretName = fmt.Sprintf("%s-%s-client-cert", addOnName, strings.ReplaceAll(item.SignerName, "/", "-"))
+		}
+
+		data, err := json.Marshal(item)
+		if err != nil {
+			return nil, err
+		}
+		h := sha256.New()
+		h.Write(data)
+		item.hash = fmt.Sprintf("%x", h.Sum(nil))
+		configs[item.hash] = item
+	}
+
+	return configs, nil
+}
+
+func defaultCommonName(clusterName, agentName, addonName string) string {
+	return fmt.Sprintf("%s:agent:%s", defaultOrganization(clusterName, addonName), agentName)
+}
+
+func defaultOrganization(clusterName, addonName string) string {
+	return fmt.Sprintf("system:open-cluster-management:cluster:%s:addon:%s", clusterName, addonName)
 }
--- a/pkg/spoke/addon/registration_controller.go
+++ b/pkg/spoke/addon/registration_controller.go
@@ -1,0 +1,251 @@
+package addon
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorhelpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+	certificatesv1beta1 "k8s.io/api/certificates/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/informers"
+	certificatesinformers "k8s.io/client-go/informers/certificates/v1beta1"
+	"k8s.io/client-go/kubernetes"
+	csrclient "k8s.io/client-go/kubernetes/typed/certificates/v1beta1"
+	"k8s.io/klog/v2"
+
+	addoninformerv1alpha1 "github.com/open-cluster-management/api/client/addon/informers/externalversions/addon/v1alpha1"
+	addonlisterv1alpha1 "github.com/open-cluster-management/api/client/addon/listers/addon/v1alpha1"
+	"github.com/open-cluster-management/registration/pkg/clientcert"
+)
+
+// addOnRegistrationController monitors ManagedClusterAddOns on hub and starts addOn registration
+// according to the registrationConfigs read from annotations of ManagedClusterAddOns. Echo addOn
+// may have multiple registrationConfigs. A clientcert.NewClientCertificateController will be started
+// for each of them.
+type addOnRegistrationController struct {
+	clusterName    string
+	agentName      string
+	kubeconfigData []byte
+	kubeClient     kubernetes.Interface
+	hubAddOnLister addonlisterv1alpha1.ManagedClusterAddOnLister
+	hubCSRInformer certificatesinformers.CertificateSigningRequestInformer
+	hubCSRClient   csrclient.CertificateSigningRequestInterface
+	recorder       events.Recorder
+
+	// registrationConfigs maps the addon name to a map of registrationConfigs whose key is the hash of
+	// the registrationConfig
+	addOnRegistrationConfigs map[string]map[string]registrationConfig
+}
+
+// NewAddOnRegistrationController returns an instance of addOnRegistrationController
+func NewAddOnRegistrationController(
+	clusterName string,
+	agentName string,
+	kubeconfigData []byte,
+	kubeClient kubernetes.Interface,
+	hubCSRInformer certificatesinformers.CertificateSigningRequestInformer,
+	hubAddOnInformers addoninformerv1alpha1.ManagedClusterAddOnInformer,
+	hubCSRClient csrclient.CertificateSigningRequestInterface,
+	recorder events.Recorder,
+) factory.Controller {
+	c := &addOnRegistrationController{
+		clusterName:              clusterName,
+		agentName:                agentName,
+		kubeconfigData:           kubeconfigData,
+		kubeClient:               kubeClient,
+		hubAddOnLister:           hubAddOnInformers.Lister(),
+		hubCSRInformer:           hubCSRInformer,
+		hubCSRClient:             hubCSRClient,
+		recorder:                 recorder,
+		addOnRegistrationConfigs: map[string]map[string]registrationConfig{},
+	}
+
+	return factory.New().
+		WithInformersQueueKeyFunc(
+			func(obj runtime.Object) string {
+				accessor, _ := meta.Accessor(obj)
+				return accessor.GetName()
+			},
+			hubAddOnInformers.Informer()).
+		WithSync(c.sync).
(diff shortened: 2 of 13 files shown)


pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,122 @@
...
+	agentName      string
+	kubeconfigData []byte
+	kubeClient     kubernetes.Interface
+	hubAddOnLister addonlisterv1alpha1.ManagedClusterAddOnLister
+	hubCSRInformer certificatesinformers.CertificateSigningRequestInformer
+	hubCSRClient   csrclient.CertificateSigningRequestInterface
+	recorder       events.Recorder
+
+	// addOnCertificateManagers maps the addon name to addOnCertificateManager
+	addOnCertificateManagers map[string]*addOnCertificateManager
```
qiujian16: why do we have another layer of manager instead of having all clientcertcontroller managed here.

I would suggest we have 
```go
addOnCertificateManagers  map[string][registrationConfig]struct{manager: *clientcertManager, stopFunc: context.CancelFunc}.
```

So in each reconcile, you are able to build  the current []registrationConfig. And compare with the addOnCertificateManagers[addonName]

I think it is fine that we do not share secret informer in an addon as the first step. We can consider having a per ns secret informer since in our case, multiple addons are installed in one ns.

pkg/spoke/addon/certificate_manager.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// no work if addon registation is not enabled
+	if len(m.Registrations) == 0 {
+		return nil
+	}
+
+	// start a shared informer factory
+	ctx, m.stopFunc = context.WithCancel(ctx)
+	kubeInformerFactory := informers.NewSharedInformerFactoryWithOptions(m.kubeClient, 10*time.Minute, informers.WithNamespace(m.InstallationNamespace))
+	go kubeInformerFactory.Start(ctx.Done())
```
qiujian16: this line should be put  after controller is started

pkg/spoke/addon/certificate_manager.go
```diff
@@ -0,0 +1,160 @@
...
+	return nil
+}
+
+// stop stops controllers started and remove resources created.
+func (m *addOnCertificateManager) stop() error {
+	if m.stopFunc == nil {
+		return nil
+	}
+
+	m.stopFunc()
```
qiujian16: I think you want to set stopFun here to nil

pkg/spoke/addon/certificate_manager.go
```diff
@@ -0,0 +1,160 @@
...
+
+// syncAddOnConfig compares the old and new addon configs, and starts/restarts controllers if necessary
+func (m *addOnCertificateManager) syncAddOnConfig(config addOnConfig, ctx context.Context) error {
+	// no work if there is no change on addon config
+	if reflect.DeepEqual(m.addOnConfig, config) {
+		return nil
+	}
+
+	// otherwise stop the controllers started
+	if err := m.stop(); err != nil {
```
qiujian16: this will be a problem since if the user add another registration, it will cause all existing secret deleted and re-register.

pkg/spoke/addon/certificate_manager.go
```diff
@@ -0,0 +1,160 @@
...
+		csrOption := clientcert.CSROption{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: fmt.Sprintf("addon-%s-%s-", m.clusterName, m.AddOnName),
+				Labels: map[string]string{
+					// the labels are only hints. Anyone could set/modify them.
+					clientcert.ClusterNameLabel: m.clusterName,
+					clientcert.AddonNameLabel:   m.AddOnName,
+				},
+			},
+			Subject: item.Subject.x509Subject(map[string]string{
```
qiujian16: lets do not consider rendering as the first step.

pkg/spoke/addon/certificate_manager.go
```diff
@@ -0,0 +1,160 @@
...
+				clusterNamePlaceHolder: m.clusterName,
+				agentNamePlaceHolder:   m.agentName,
+			}),
+			SignerName:      item.SignerName,
+			EventFilterFunc: createCSREventFilterFunc(m.clusterName, m.AddOnName, item.SignerName),
+		}
+
+		controllerName := fmt.Sprintf("ClientCertController@addon:%s", m.AddOnName)
+		if item.SignerName != certificatesv1beta1.KubeAPIServerClientSignerName {
```
qiujian16: why we need this if?

pkg/spoke/addon/configuration.go
```diff
@@ -0,0 +1,123 @@
...
+	registrationAnnotation          = "registrations"
+	clusterNamePlaceHolder          = "{ClusterName}"
+	agentNamePlaceHolder            = "{AgentName}"
+)
+
+// addOnConfig contains meta information of an addon
+type addOnConfig struct {
+	AddOnName             string
+	InstallationNamespace string
+	BootstrapSecretName   string
```
qiujian16: let's remove bootstrap secret as the first step

pkg/spoke/spokeagent.go
```diff
@@ -289,11 +292,36 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
...
+		)
+
+	}
+
 	go hubKubeInformerFactory.Start(ctx.Done())
 	go hubClusterInformerFactory.Start(ctx.Done())
 	go spokeKubeInformerFactory.Start(ctx.Done())
 	go namespacedSpokeKubeInformerFactory.Start(ctx.Done())
 	go spokeClusterInformers.Start(ctx.Done())
+	if features.DefaultMutableFeatureGate.Enabled(features.AddonManagement) {
```
qiujian16: you can always start it, will be no-op if the feature gate is not enabled

test/integration/addon_registration_test.go
```diff
@@ -0,0 +1,234 @@
...
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		ginkgo.By("Create Addon cr with required annotations")
+		addOn := &addonv1alpha1.ManagedClusterAddOn{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      addOnName,
+				Namespace: managedClusterName,
+				Annotations: map[string]string{
+					"installNamespace": addOnName,
+					"registrations":    `[{"signerName":"kubernetes.io/kube-apiserver-client"}]`,
```
qiujian16: we would also need a case for customized signer.

pkg/spoke/addon/registration_controller.go
elgnay: I have to use two maps below instead the one you suggested.
```golang
	// registrationConfigs maps the addon name to a slice of registrationConfigs
	addOnRegistrationConfigs map[string][]registrationConfig
	// stopRegistrationFuncs maps the key of a registrationConfig to its stopRegistrationFunc
	stopRegistrationFuncs map[string]context.CancelFunc
```
`registrationConfig` struct is a invalid type for map key because registrationConfig.Subject.Organization is an array, which is not comparable.

pkg/spoke/addon/configuration.go
```diff
@@ -0,0 +1,159 @@
...
+
+// addOnConfig contains meta information of an addon
+type addOnConfig struct {
+	AddOnName             string
+	InstallationNamespace string
+}
+
+// registrationConfig contains necessary information for addon registration
+type registrationConfig struct {
+	AddOnName             string      `json:"addOnName,omitempty"`
```
deads2k: why have public names on internal types?

pkg/spoke/addon/configuration.go
```diff
@@ -0,0 +1,159 @@
...
+	InstallationNamespace string      `json:"installNamespace,omitempty"`
```
deads2k: why are there json tags here?  In kube we try to separate serialized types from unserialized types.  We do this because serialized types represent a guarantee to a user. Unserialized types have no guarantees.

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,232 @@
...
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		delete(c.stopRegistrationFuncs, config.hash)
+	}
+	return nil
+}
+
+// cleanup cleans both the registration configs and client certificate controllers for the addon
+func (c *addOnRegistrationController) cleanup(addOnName string, ctx context.Context) error {
```
deads2k: context should always be passed first

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,232 @@
...
+
+	go kubeInformerFactory.Start(ctx.Done())
+	go clientCertController.Run(ctx, 1)
+
+	c.stopRegistrationFuncs[config.hash] = stopFunc
+}
+
+// stopRegistration stops the client certificate controller for the given config
+func (c *addOnRegistrationController) stopRegistration(config registrationConfig, ctx context.Context) error {
+	if stopFunc, ok := c.stopRegistrationFuncs[config.hash]; ok {
```
deads2k: why is this registration function keyed by the config.hash instead of being keyed b y addonName?  It looks like this controller determines the config and the hashed value based on the addonname.  Having a single key would make a litlt easier to read.

EDIT: I've got to the second caller of this method.  Perhaps add a comment clarifying how an addon could have multiple configs when annotations change.

pkg/spoke/addon/configuration.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return &addOnConfig{
+		AddOnName:             addOn.Name,
+		InstallationNamespace: installationNamespace,
+	}, nil
+}
+
+// getRegistrationConfigs returns a slice of registrationConfig from annotations of a addon
+func getRegistrationConfigs(addOnName string, annotations map[string]string) ([]registrationConfig, error) {
+	ns := annotations[installationNamespaceAnnotation]
```
deads2k, line 92: if we're relying on these like an API, why are they not API fields.

pkg/spoke/addon/configuration.go
```diff
@@ -0,0 +1,159 @@
...
+	"fmt"
+	"strings"
+
+	addonv1alpha1 "github.com/open-cluster-management/api/addon/v1alpha1"
+	certificatesv1beta1 "k8s.io/api/certificates/v1beta1"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	installationNamespaceAnnotation = "installNamespace"
```
deads2k: these shoudl be namespaced in your API group.  `addon.something.something.io/installNamespace`

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,232 @@
...
+	c := &addOnRegistrationController{
+		clusterName:              clusterName,
+		agentName:                agentName,
+		kubeconfigData:           kubeconfigData,
+		kubeClient:               kubeClient,
+		hubAddOnLister:           hubAddOnInformers.Lister(),
+		hubCSRInformer:           hubCSRInformer,
+		hubCSRClient:             hubCSRClient,
+		recorder:                 recorder,
+		addOnRegistrationConfigs: map[string][]registrationConfig{},
```
deads2k: why is this a single value instead of a slice since addons can logically have multiple registrationConfigs where only one is active

pkg/spoke/addon/configuration.go
```diff
@@ -0,0 +1,159 @@
...
+	registrationAnnotation          = "registrations"
```
deads2k: can I get a comment about what is stored in this annotation and in what format?

pkg/spoke/addon/configuration.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// getRegistrationConfigs returns a slice of registrationConfig from annotations of a addon
+func getRegistrationConfigs(addOnName string, annotations map[string]string) ([]registrationConfig, error) {
+	ns := annotations[installationNamespaceAnnotation]
+	if len(ns) == 0 {
+		return nil, nil
+	}
+
+	registration := annotations[registrationAnnotation]
```
deads2k, line 97: where is this being written?

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,232 @@
...
+	"k8s.io/client-go/kubernetes"
+	csrclient "k8s.io/client-go/kubernetes/typed/certificates/v1beta1"
+	"k8s.io/klog/v2"
+
+	addoninformerv1alpha1 "github.com/open-cluster-management/api/client/addon/informers/externalversions/addon/v1alpha1"
+	addonlisterv1alpha1 "github.com/open-cluster-management/api/client/addon/listers/addon/v1alpha1"
+	"github.com/open-cluster-management/registration/pkg/clientcert"
+)
+
+// addOnRegistrationController monitors ManagedClusterAddOn crs on hub and starts addOn registration
```
deads2k: enumerate the actual controllers you're starting. The first one is a clientcert.NewClientCertificateController.

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,232 @@
...
+
+	clientCertController := clientcert.NewClientCertificateController(
+		clientCertOption,
+		csrOption,
+		c.hubCSRInformer,
+		c.hubCSRClient,
+		kubeInformerFactory.Core().V1().Secrets(),
+		c.kubeClient.CoreV1(),
+		c.recorder,
+		fmt.Sprintf("ClientCertController@addon:%s:signer:%s", config.AddOnName, config.SignerName),
```
deads2k: nit: assign this to a named variable above, so I have something like `whateverThisFieldIs := fmt.Sprintf("ClientCertController@addon:%s:signer:%s", config.AddOnName, config.SignerName)` just to help future us

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,232 @@
...
+	)
+
+	go kubeInformerFactory.Start(ctx.Done())
+	go clientCertController.Run(ctx, 1)
+
+	c.stopRegistrationFuncs[config.hash] = stopFunc
```
deads2k: I would prefer to return the stop function and have `c.stopRegistrationFuncs[config.hash] = stopFunc` in the caller.

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,232 @@
...
+		if labels[clientcert.ClusterNameLabel] != clusterName {
+			return false
+		}
+		// only enqueue csr created for a specific addon
+		if labels[clientcert.AddonNameLabel] != addOnName {
+			return false
+		}
+
+		// only enqueue csr with a specific signer name
+		csr, ok := obj.(*certificatesv1beta1.CertificateSigningRequest)
```
deads2k, line 239: you should update to csr v1.  Is anything stopping you from doing that nowish in a followup?

deads2k:
This is a solid start.  I'd like a second review after the comments are addressed.  Please get me on slack then.

pkg/spoke/addon/registration_controller.go
elgnay: Updated.

pkg/spoke/addon/registration_controller.go
elgnay: An addon may have more than one registrationConfig and all of them are active. That's the reason why a registrationConfig slice is used as the type of value.

pkg/spoke/addon/configuration.go
elgnay: Doc added for registrationAnnotation.

pkg/spoke/addon/configuration.go
elgnay: The addOn controller running on hub will set this annotation on addOn CR.

pkg/spoke/addon/registration_controller.go
elgnay: Added in comments.

pkg/spoke/addon/registration_controller.go
elgnay: Code updated.

pkg/spoke/addon/registration_controller.go
elgnay: Updated as  suggested.

pkg/spoke/addon/registration_controller.go
elgnay: I'll do it in a separated task.

pkg/spoke/addon/registration_controller.go
elgnay: An addOn may have multiple registrationConfigs with different signers. Each registrationConfig will have a clientcert.NewClientCertificateController started.

pkg/spoke/addon/configuration.go
elgnay: It has to be public because I want to add json tag on it.

pkg/spoke/addon/configuration.go
elgnay: Json tags are added on fields because,
- I need them to unserialized registrationConfs from annotation of `ManagedClusterAddOn`.
- I serialize `registrationConf` into json string and then calculate the hash of `registrationConf`. `registrationConfig` is not comparable because `registrationConfig.Subject.Organization` is an array. Hash is used to compare registrationConfigs and also can be used as key of map.

pkg/spoke/addon/configuration.go
elgnay: Updated.

pkg/spoke/addon/configuration.go
elgnay: We will add them into api in another pr. https://github.com/open-cluster-management/api/pull/73

pkg/spoke/addon/configuration.go
```diff
@@ -16,6 +27,48 @@ type addOnConfig struct {
 	InstallationNamespace string
 }
 
+// registrationConfig contains necessary information for addon registration
```
qiujian16, line 31: Add a TODO to refactor here when we update api spec.

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,250 @@
...
+		WithSync(c.sync).
+		ToController("AddOnRegistrationController", recorder)
+}
+
+func (c *addOnRegistrationController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	addOnName := syncCtx.QueueKey()
+	klog.V(4).Infof("Reconciling addOn %q", addOnName)
+
+	addOn, err := c.hubAddOnLister.ManagedClusterAddOns(c.clusterName).Get(addOnName)
+	if errors.IsNotFound(err) {
```
qiujian16, line 85: I think we need to add a finalizer to handle deletion?

pkg/spoke/addon/registration_controller.go
```diff
@@ -0,0 +1,250 @@
...
+
+	for _, config := range configs {
+		cachedConfig, ok := cachedConfigs[config.hash]
+		switch {
+		case ok:
+			// get the stopFunc from the cached registration configs
+			config.stopFunc = cachedConfig.stopFunc
+		default:
+			// start registration for the new added configs
+			config.stopFunc = c.startRegistration(ctx, config)
```
qiujian16: why not just put configs[config.hash] = config in the if block?

```go
if !ok {
        // start registration for the new added configs
	config.stopFunc = c.startRegistration(ctx, config)
        configs[config.hash] = config
}
```

elgnay: Rewrote in a similar way to remove the switch clause.

pkg/spoke/addon/configuration.go
elgnay: Added.

pkg/spoke/addon/registration_controller.go
elgnay: If we add a finalizer on `ManagedClusterAddOn` cr on hub, It might impact the detach of the managed cluster.

elgnay: I added a TODO and will create a followup task to add finalizer.

pkg/spoke/addon/configuration.go
deads2k: > The addOn controller running on hub will set this annotation on addOn CR.

This looks like it should be a field on status to me so you can have the structured shape.

deads2k:
/approve

I like the direction this is moving.  I'd like the json annotation promoted into the API, but I'm ok pushing this through if you work on getting it into status.

I'll leave lgtm  with 

/assign @qiujian16

PR #125, opened 2021-04-06
upgrade CRD to v1 and k8s lib to v0.21.0-rc.0

zhiweiyin318:
1. ~upgrade webhook to v1~ (openshift/generic-admission-server does not support webhook v1 yet)
2. upgrade CRD to v1 in deploy
3. upgrade k8s lib to v0.21.0-rc.0
4. upgrade kubebuilder to v1.19.2


qiujian16:
also need to change here https://github.com/open-cluster-management/registration/blob/main/test/e2e/loopback_test.go#L46

zhiweiyin318:
> also need to change here https://github.com/open-cluster-management/registration/blob/main/test/e2e/loopback_test.go#L46

have changed to v1

qiujian16 (commented):
/approve
/lgtm

PR #126, opened 2021-04-08
Fix invalid client certificate issue

elgnay:

### DIFF
--- a/pkg/clientcert/certificate.go
+++ b/pkg/clientcert/certificate.go
@@ -4,7 +4,6 @@ import (
 	"crypto/x509/pkix"
 	"errors"
 	"fmt"
-	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/clientcert/certificate.go
```diff
@@ -80,16 +80,18 @@ func IsCertificateValid(certData []byte, subject *pkix.Name) (bool, error) {
...
 	// check subject of certificates
+	organizationSets := sets.NewString(subject.Organization...)
+	organizationalUnitSets := sets.NewString(subject.OrganizationalUnit...)
 	for _, cert := range certs {
 		if cert.Subject.CommonName != subject.CommonName {
 			continue
 		}
 
-		if !reflect.DeepEqual(cert.Subject.Organization, subject.Organization) {
+		if !organizationSets.Equal(sets.NewString(cert.Subject.Organization...)) {
```
qiujian16: i would want removing the comparison of o/ou now. We can add it when we think it is necessary.

qiujian16 (commented):
/approve

pkg/clientcert/certificate.go
elgnay: Removed.

skeeey:
refer to https://github.com/open-cluster-management/backlog/issues/11314

PR #128, opened 2021-04-08
compatibility - manage addon lease

skeeey:

### DIFF
--- a/pkg/spoke/addon/lease_controller.go
+++ b/pkg/spoke/addon/lease_controller.go
@@ -2,6 +2,7 @@ package addon
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	addonv1alpha1 "github.com/open-cluster-management/api/addon/v1alpha1"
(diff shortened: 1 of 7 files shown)


pkg/spoke/addon/lease_controller.go
```diff
@@ -103,12 +108,39 @@ func (c *managedClusterAddOnLeaseController) syncSingle(ctx context.Context,
...
 	recorder events.Recorder) error {
+	now := c.clock.Now()
+	gracePeriod := time.Duration(leaseDurationTimes*leaseDurationSeconds) * time.Second
 	// addon lease name should be same with the addon name.
 	observedLease, err := c.leaseLister.Leases(leaseNamespace).Get(addOn.Name)
 
 	var condition metav1.Condition
 	switch {
 	case errors.IsNotFound(err):
+		// for backward compatible, before release-2.3, addons update their leases on hub cluster,
```
qiujian16, line 119: this will double the lease call with 2.3 hub and 2.2 agent. FYI @TheRealHaoLiu but i think we need this to support version skew. I think we will not need klusterlet-addon-controller to maintain status of managedclusteraddon anymore, right?

@hanqiuzh @itdove @leena-jawale

hanqiuzh: klusterlet-addon-controller also has a progressing=true when installing/creating the manifestwork, but for type available, I think we will not touch going forward.

PR #129, opened 2021-04-12
update addon api

skeeey:


pkg/spoke/addon/configuration.go
```diff
@@ -12,30 +12,14 @@ import (
...
 
 // registrationConfig contains necessary information for addon registration
 // TODO: Refactor the code here once the registration configuration is available in spec of ManagedClusterAddOn
 type registrationConfig struct {
-	AddOnName             string `json:"addOnName,omitempty"`
-	InstallationNamespace string `json:"installNamespace,omitempty"`
-	// SignerName is used to create csr. The default value is "kubernetes.io/kube-apiserver-client".
-	SignerName string      `json:"signerName,omitempty"`
-	Subject    certSubject `json:"subject,omitempty"`
+	AddOnName             string                           `json:"addOnName,omitempty"`
```
qiujian16: Do not need json part, and it does not need to be public

skeeey: updated, using registration hash instead of registrationConfig hash

pkg/spoke/addon/configuration.go
```diff
@@ -46,91 +30,69 @@ type registrationConfig struct {
...
+	return installationNamespace
 }
 
 // getRegistrationConfigs reads annotations of a addon and returns a map of registrationConfig whose
 // key is the hash of the registrationConfig
-func getRegistrationConfigs(addOnName string, annotations map[string]string) (map[string]registrationConfig, error) {
-	ns := annotations[installationNamespaceAnnotation]
-	if len(ns) == 0 {
+func getRegistrationConfigs(addOn *addonv1alpha1.ManagedClusterAddOn) (map[string]registrationConfig, error) {
+	if addOn.Status.Registrations == nil || len(addOn.Status.Registrations) == 0 {
```
qiujian16: I think you can use len(addOn.Status.Registrations) == 0 only

skeeey: updated

pkg/spoke/addon/configuration.go
```diff
@@ -46,91 +30,69 @@ type registrationConfig struct {
...
+	return installationNamespace
 }
 
 // getRegistrationConfigs reads annotations of a addon and returns a map of registrationConfig whose
 // key is the hash of the registrationConfig
-func getRegistrationConfigs(addOnName string, annotations map[string]string) (map[string]registrationConfig, error) {
-	ns := annotations[installationNamespaceAnnotation]
-	if len(ns) == 0 {
+func getRegistrationConfigs(addOn *addonv1alpha1.ManagedClusterAddOn) (map[string]registrationConfig, error) {
+	if len(addOn.Status.Registrations) == 0 {
```
deads2k: nit: I think this is unnecessary, wince without registrations, that loop doesn't run and you return empty.

deads2k:
just a nit.

/approve

pkg/spoke/addon/configuration.go
skeeey: yeah, this is indeed unnecessary, removed

PR #130, opened 2021-04-13
Support DNS names of client cert

elgnay:

### DIFF
--- a/pkg/clientcert/cert_controller.go
+++ b/pkg/clientcert/cert_controller.go
@@ -6,7 +6,6 @@ import (
 	"crypto/x509/pkix"
 	"fmt"
 	"math/rand"
-	"reflect"
 	"time"
 
 	"github.com/openshift/library-go/pkg/controller/factory"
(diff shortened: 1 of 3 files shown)


pkg/clientcert/cert_controller.go
```diff
@@ -183,25 +185,16 @@ func (c *clientCertificateController) sync(ctx context.Context, syncCtx factory.
...
-			return err
-		}
-	}
-
 	// create a csr to request new client certificate if
 	// a. there is no valid client certificate issued for the current cluster/agent
 	// b. client certificate exists and has less than a random percentage range from 20% to 25% of its life remaining
 	if c.hasValidClientCertificate(secret) {
+		// add additional data into client certificate secret
+		if err := c.appendAdditinalData(secret); err != nil {
```
qiujian16: Additional

pkg/clientcert/cert_controller.go
```diff
@@ -183,25 +185,16 @@ func (c *clientCertificateController) sync(ctx context.Context, syncCtx factory.
...
+			return err
+		}
```
qiujian16: since we save secret here https://github.com/open-cluster-management/registration/pull/130/files#diff-710e745dc0a42d985a34b2b2b79aaa8d5d559756ad0f5047f1c6922cbf8beb7dL179, why do we need to do  it again here?

elgnay: Here is to save the additional data, like agent name which might be missing and recovered from cert

pkg/clientcert/cert_controller.go
elgnay: Updated

qiujian16 (commented):
/approve
/lgtm

PR #135, opened 2021-04-16
run as non-root

zhiweiyin318:
Signed-off-by: Zhiwei Yin <zyin@redhat.com>


qiujian16:
also needs to change the deployment spec so we can test in e2e

PR #136, opened 2021-04-19
include cluster name in name of client cert controller

elgnay:


pkg/spoke/spokeagent.go
```diff
@@ -172,6 +172,7 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
 			return err
 		}
 
+		controllerName := fmt.Sprintf("BootstrapClientCertController@cluster:%s", o.ClusterName)
```
qiujian16, line 175: I  think this line should be out of if? What if the agent is bootstrapped and restarted?

PR #138, opened 2021-04-21
add resync logic for addon

elgnay:

### DIFF
--- a/pkg/spoke/addon/registration_controller.go
+++ b/pkg/spoke/addon/registration_controller.go
@@ -38,6 +38,8 @@ type addOnRegistrationController struct {
 	hubCSRClient   csrclient.CertificateSigningRequestInterface
 	recorder       events.Recorder
 
+	startRegistrationFunc func(ctx context.Context, config registrationConfig) context.CancelFunc
+
 	// registrationConfigs maps the addon name to a map of registrationConfigs whose key is the hash of
 	// the registrationConfig
 	addOnRegistrationConfigs map[string]map[string]registrationConfig
@@ -66,6 +68,8 @@ func NewAddOnRegistrationController(
 		addOnRegistrationConfigs: map[string]map[string]registrationConfig{},
 	}
 
+	c.startRegistrationFunc = c.startRegistration
+
 	return factory.New().
 		WithInformersQueueKeyFunc(
 			func(obj runtime.Object) string {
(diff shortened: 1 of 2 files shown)


pkg/spoke/addon/registration_controller.go
```diff
@@ -74,11 +79,44 @@ func NewAddOnRegistrationController(
...
+
+	// handle resync
+	addOns, err := c.hubAddOnLister.ManagedClusterAddOns(c.clusterName).List(labels.Everything())
+	if err != nil {
+		return err
+	}
+
+	errs := []error{}
+	for _, addOn := range addOns {
+		_, err := c.hubAddOnClient.AddonV1alpha1().ManagedClusterAddOns(c.clusterName).Get(ctx, addOn.Name, metav1.GetOptions{})
```
qiujian16: why not get from cache?

pkg/spoke/addon/registration_controller.go
```diff
@@ -74,11 +79,44 @@ func NewAddOnRegistrationController(
...
 func (c *addOnRegistrationController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
-	addOnName := syncCtx.QueueKey()
+	queueKey := syncCtx.QueueKey()
+	if queueKey != factory.DefaultQueueKey {
+		// sync a particular addOn
+		return c.syncAddOn(ctx, syncCtx, queueKey)
+	}
+
+	// handle resync
+	addOns, err := c.hubAddOnLister.ManagedClusterAddOns(c.clusterName).List(labels.Everything())
```
qiujian16: i think we get all existing addons from addOnRegistrationConfigs,  so we can compare with current lister

qiujian16 (commented):
we  need some test for this

pkg/spoke/addon/registration_controller.go
elgnay: Code updated to sync between cached registration configs and ManagedClusterAddOns in informer cache.

pkg/spoke/addon/registration_controller.go
elgnay: Code updated to sync between cached registration configs and ManagedClusterAddOns in informer cache.

elgnay:
> we need some test for this

Unit test case added.

PR #140, opened 2021-04-22
reset csr name only if csr sync failed or secret saved

elgnay:
The related issue: https://github.com/open-cluster-management/backlog/issues/11821


qiujian16 (commented):
/approve
/lgtm

PR #142, opened 2021-04-30
change to golang version 1.16

xuezhaojun:
Signed-off-by: xuezhaojun <xuezhaojun@users.noreply.github.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/open-cluster-management/registration
 
-go 1.15
+go 1.16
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/openshift/origin-release:golang-1.15 AS builder
+FROM registry.ci.openshift.org/open-cluster-management/builder:go1.16-linux AS builder
```
qiujian16: i would not change this since  it will avoid community  user to build the image. If we have a quota issue on docker hub, we may consider using golang image in gcr.io

PR #148, opened 2021-05-27
update crds

zhiweiyin318:
Signed-off-by: Zhiwei Yin <zyin@redhat.com>


hack/copy-crds.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+cp ./vendor/github.com/open-cluster-management/api/cluster/v1/0000_00_clusters.open-cluster-management.io_managedclusters.crd.yaml ./deploy/hub/managedcluster.crd.yaml
```
qiujian16: if we add this, we need to add script to verify also

zhiweiyin318: done

PR #150, opened 2021-06-17
keep compatibility with lease API

skeeey:

### DIFF
--- a/pkg/spoke/addon/lease_controller.go
+++ b/pkg/spoke/addon/lease_controller.go
@@ -20,9 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/clock"
-	coordinformers "k8s.io/client-go/informers/coordination/v1"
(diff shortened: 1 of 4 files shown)


pkg/spoke/addon/lease_controller.go
```diff
@@ -59,8 +59,11 @@ func NewManagedClusterAddOnLeaseController(clusterName string,
 		hubLeaseClient: hubLeaseClient,
 		leaseLister:    leaseInformer.Lister(),
 	}
+
+	// TODO because kube lease API is stable after kube v1.17, to keep compatibility, we does not
+	// register the lease informer here (if kube version is lower than v1.17, using lease informer
+	// will result in this controller failed), in the future, we will register lease informer here
```
qiujian16: We do not add leaser informer to support kubernetes version lower than 1.17.  Lease v1 api is introduced in  v1.17, hence adding lease informer in this controller will cause the hang of informer cache sync and result in fatal exit of this controller. The code will be factored when we no longer support kubernetes version lower than 1.17.

skeeey: updated

qiujian16 (commented):
/approve
/lgtm

PR #151, opened 2021-06-18
Fix issues in unit-test and e2e-test

xuezhaojun:
### Unit-test
function `getCertValidityPeriod` will return `NotBefore` and `NotAfter` with minimal time scope. 
In this test, first_cert and second_cert is like below:
```
first cert       : | time.Now() --------- 10 secs ------- NotAfter | 
second cert :            | time.Now() -- 5 secs -- NotAfter | 
```
Which means **normally** we get a cert with `NotBefore` and `NotAfter` equals to second cert.

But if CPU resource is limited, and all unit-test cases run concurrently, second cert may created after first cert created by several seconds, which like below:
```
first cert       : | time.Now() --------- 10 secs ------- NotAfter | 
second cert :                                                 | time.Now() -- 5 secs -- NotAfter | 
```

So we need to do comparison at first. 

### e2e-test
` NotFound` is a special case and should not return err.

Signed-off-by: xuezhaojun <zxue@redhat.com>

### DIFF
--- a/pkg/clientcert/certificate_test.go
+++ b/pkg/clientcert/certificate_test.go
@@ -1,7 +1,6 @@
 package clientcert
 
 import (
-	"crypto/x509"
 	"crypto/x509/pkix"
 	"testing"
 	"time"
@@ -161,15 +160,21 @@ func TestIsCertificateValid(t *testing.T) {
 
 func TestGetCertValidityPeriod(t *testing.T) {
 	certs := []byte{}
-	certs = append(certs, testinghelpers.NewTestCert("cluster0", 10*time.Second).Cert...)
-	secondCert := testinghelpers.NewTestCert("cluster0", 5*time.Second).Cert
+	firstCert := testinghelpers.NewTestCert("cluster0", 5*time.Second).Cert
+	certs = append(certs, firstCert...)
+	secondCert := testinghelpers.NewTestCert("cluster0", 10*time.Second).Cert
 	certs = append(certs, secondCert...)
-	expectedCerts, _ := certutil.ParseCertsPEM(secondCert)
+	firstCerts, _ := certutil.ParseCertsPEM(firstCert)
+	secondCerts, _ := certutil.ParseCertsPEM(secondCert)
+	notBefore := secondCerts[0].NotBefore
+	notAfter := firstCerts[0].NotAfter
+
 	cases := []struct {
-		name         string
-		secret       *corev1.Secret
-		expectedCert *x509.Certificate
-		expectedErr  string
+		name        string
+		secret      *corev1.Secret
+		expectedErr string
+		notBefore   time.Time
+		notAfter    time.Time
 	}{
 		{
 			name:        "no data",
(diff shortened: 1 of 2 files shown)


test/e2e/loopback_test.go
```diff
@@ -308,6 +308,9 @@ var _ = ginkgo.Describe("Loopback registration [development]", func() {
 			var err error
 			managedCluster, err = managedClusters.Get(context.TODO(), clusterName, metav1.GetOptions{})
 			if err != nil {
+				if err.(*errors.StatusError).ErrStatus.Reason == metav1.StatusReasonNotFound {
```
skeeey: you can use `k8s.io/apimachinery/pkg/api/errors` to judge not found, e.g.

```
        if errors.IsNotFound(err) {
		return false, nil
	}
	if err != nil {
		return false, err
	}
```

pkg/clientcert/certificate_test.go
```diff
@@ -161,10 +161,24 @@ func TestIsCertificateValid(t *testing.T) {
...
 func TestGetCertValidityPeriod(t *testing.T) {
 	certs := []byte{}
-	certs = append(certs, testinghelpers.NewTestCert("cluster0", 10*time.Second).Cert...)
+	firstCert := testinghelpers.NewTestCert("cluster0", 10*time.Second).Cert
+	certs = append(certs, firstCert...)
 	secondCert := testinghelpers.NewTestCert("cluster0", 5*time.Second).Cert
 	certs = append(certs, secondCert...)
-	expectedCerts, _ := certutil.ParseCertsPEM(secondCert)
+
+	firstCerts,_:= certutil.ParseCertsPEM(firstCert)
```
skeeey: keep a space between `firstCerts` and `_`

pkg/clientcert/certificate_test.go
```diff
@@ -161,10 +161,24 @@ func TestIsCertificateValid(t *testing.T) {
...
+	secondCerts,_ := certutil.ParseCertsPEM(secondCert)
```
skeeey: keep a space between `secondCerts` and  `_`

pkg/clientcert/certificate_test.go
```diff
@@ -161,10 +161,24 @@ func TestIsCertificateValid(t *testing.T) {
...
+
+	fc := firstCerts[0]
+	sc := secondCerts[0]
+
+	var expectedCert *x509.Certificate
+	if fc.NotAfter.Before(sc.NotAfter) {
```
skeeey: this is confusion, could we change the sequence of `firstCert` and `secondCert` to make the `firstCert` always has a short duration, then use it as the `expectedCert`

test/e2e/loopback_test.go
xuezhaojun: Done

pkg/clientcert/certificate_test.go
xuezhaojun: Done

pkg/clientcert/certificate_test.go
xuezhaojun: Yes, and I guess in this way we would have less code

pkg/clientcert/certificate_test.go
skeeey: should be
```
firstCerts, _ := certutil.ParseCertsPEM(firstCert)
secondCert, _ := certutil.ParseCertsPEM(secondCert)
```

pkg/clientcert/certificate_test.go
```diff
@@ -187,23 +204,24 @@ func TestGetCertValidityPeriod(t *testing.T) {
...
+			notBefore: notBefore,
+			notAfter:  notAfter,
 		},
 	}
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
 			notBefore, notAfter, err := getCertValidityPeriod(c.secret)
 			testinghelpers.AssertError(t, err, c.expectedErr)
-			if c.expectedCert == nil {
+			if c.expectedErr != "" {
```
skeeey: I think you should check this error

pkg/clientcert/certificate_test.go
```diff
@@ -161,15 +160,33 @@ func TestIsCertificateValid(t *testing.T) {
...
+	certs = append(certs, firstCert...)
 	secondCert := testinghelpers.NewTestCert("cluster0", 5*time.Second).Cert
 	certs = append(certs, secondCert...)
-	expectedCerts, _ := certutil.ParseCertsPEM(secondCert)
+	firstCerts, _ := certutil.ParseCertsPEM(firstCert)
+	secondCerts, _ := certutil.ParseCertsPEM(secondCert)
+	fC := firstCerts[0]
+	sC := secondCerts[0]
+
+	var notBefore, notAfter time.Time
```
skeeey: can we eliminate this judgement? like this

```
firstCert := testinghelpers.NewTestCert("cluster0", 5*time.Second).Cert
secondCert := testinghelpers.NewTestCert("cluster0", 10*time.Second).Cert

notBefore := secondCert. NotBefore
notAfter := firstCert.NotAfter
```

pkg/clientcert/certificate_test.go
skeeey:
```
if err != nil {
return
}
```

PR #152, opened 2021-06-21
Change KUSTOMIZE related path to absolute path

xuezhaojun:
Signed-off-by: xuezhao <zxue@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,8 @@ IMAGE_TAG?=latest
 IMAGE_NAME?=$(IMAGE_REGISTRY)/registration:$(IMAGE_TAG)
 KUBECONFIG ?= ./.kubeconfig
 KUBECTL?=kubectl
-KUSTOMIZE?=$(PERMANENT_TMP_GOPATH)/bin/kustomize
+PWD=$(shell pwd)
+KUSTOMIZE?=$(PWD)/$(PERMANENT_TMP_GOPATH)/bin/kustomize
 KUSTOMIZE_VERSION?=v3.5.4
 KUSTOMIZE_ARCHIVE_NAME?=kustomize_$(KUSTOMIZE_VERSION)_$(GOHOSTOS)_$(GOHOSTARCH).tar.gz
 kustomize_dir:=$(dir $(KUSTOMIZE))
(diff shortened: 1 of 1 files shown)


qiujian16:
why it must be absolute path?

skeeey:
If user already has kustomize in their env, and use KUSTOMIZE to set their kustomize, the make will be failed

xuezhaojun:
> why it must be absolute path?

In our makefile, we also provide users an option to use an existing kustomize, and build will failed if we use related path.
```mk
ensure-kustomize:
ifeq "" "$(wildcard $(KUSTOMIZE))"
	$(info Installing kustomize into '$(KUSTOMIZE)')
	mkdir -p '$(kustomize_dir)'
	curl -s -f -L https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2F$(KUSTOMIZE_VERSION)/$(KUSTOMIZE_ARCHIVE_NAME) -o '$(kustomize_dir)$(KUSTOMIZE_ARCHIVE_NAME)'
	tar -C '$(kustomize_dir)' -zvxf '$(kustomize_dir)$(KUSTOMIZE_ARCHIVE_NAME)'
	chmod +x '$(KUSTOMIZE)';
else
	$(info Using existing kustomize from "$(KUSTOMIZE)"). #  Option to use an existing kustomize
endif
```

qiujian16:
if a user has a kustomize binary in `/usr/local/bin`. How does this fix help?

skeeey:
```
KUSTOMIZE?=$(PERMANENT_TMP_GOPATH)/bin/kustomize
```
if user does not define the `KUSTOMIZE`, the `$(PERMANENT_TMP_GOPATH)/bin/kustomize` will be used, even the kustomize is in the  user `PATH`, so current implement is that if user want to use his own kustomize, he must specify the `KUSTOMIZE`

qiujian16:
I see. I think we need to fix it in other repos also.

/approve
/lgtm

xuezhaojun:
> I see. I think we need to fix it in other repos also.
> 
> /approve
> /lgtm

Yes, I am working on repo work and addon-framework about the same issue.

PR #159, opened 2021-08-06
update finalizer

huiwq1990:
use controller-runtime to delete or update finalizer.

### DIFF
--- a/pkg/hub/managedcluster/controller.go
+++ b/pkg/hub/managedcluster/controller.go
@@ -9,6 +9,7 @@ import (
 	listerv1 "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
 	v1 "open-cluster-management.io/api/cluster/v1"
 	"open-cluster-management.io/registration/pkg/helpers"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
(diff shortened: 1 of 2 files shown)


pkg/hub/managedcluster/controller.go
```diff
@@ -78,26 +79,26 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
...
 	}
 
 	// Spoke cluster is deleting, we remove its related resources
 	if !managedCluster.DeletionTimestamp.IsZero() {
 		if err := c.removeManagedClusterResources(ctx, managedClusterName); err != nil {
 			return err
 		}
-		return c.removeManagedClusterFinalizer(ctx, managedCluster)
+		if controllerutil.ContainsFinalizer(managedCluster, managedClusterFinalizer) {
+			controllerutil.RemoveFinalizer(managedCluster, managedClusterFinalizer)
```
qiujian16, line 96: I hasn't take a look at the details of RemoveFinalizer, will it ensure the ordering after a deletion?

huiwq1990: Sorry, I don't understand the ordering meaing, did you say the order of finalizes? Why we need order? `RemoveFinalizer` will not change finalize order.

```
			f = append(f[:i], f[i+1:]...)

```

qiujian16: yes, the ordering of the finalizers. Some controllers may rely on the ordering of the finalizers, it is not normal but it happens.

huiwq1990: RemoveFinalizer will keep the order. By the way, could you specify which controller use this feature?

qiujian16: not in registration, but an external consumer (controller) could do that.

qiujian16:
This may need to run `go mod tidy` &  `go mod vendor` to update the deps. The problem we met with controller-runtime is its inconsistent deps of kube related library with library-go... I think that is one of the main reason we did not use controller-runtime utility much before.

PR #163, opened 2021-08-17
add the missed step and correct output message

xauthulei:
Add the missed step when cloning source and correct the output message of example. Thanks


xauthulei:
@qiujian16 , Thanks

PR #164, opened 2021-08-23
Add controller to generate addon label on cluster

elgnay:
The related issue: https://github.com/open-cluster-management/backlog/issues/15279
Enhancement proposal: https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/17-cluster-features

### DIFF
--- a/pkg/hub/addon/discovery_controller.go
+++ b/pkg/hub/addon/discovery_controller.go
@@ -1,0 +1,225 @@
+package addon
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	addonv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	addoninformerv1alpha1 "open-cluster-management.io/api/client/addon/informers/externalversions/addon/v1alpha1"
+	addonlisterv1alpha1 "open-cluster-management.io/api/client/addon/listers/addon/v1alpha1"
+	clientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	clusterv1informer "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	clusterv1listers "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+)
+
+const (
+	addOnFeaturePrefix     = "feature.open-cluster-management.io/addon-"
+	addOnStatusAvailable   = "available"
+	addOnStatusUnhealthy   = "unhealthy"
+	addOnStatusUnreachable = "unreachable"
+)
+
+// addOnFeatureDiscoveryController monitors ManagedCluster and its ManagedClusterAddOns on hub and
+// create/update/delete labels of the ManagedCluster to reflect the status of addons.
+type addOnFeatureDiscoveryController struct {
+	clusterClient clientset.Interface
+	clusterLister clusterv1listers.ManagedClusterLister
+	addOnLister   addonlisterv1alpha1.ManagedClusterAddOnLister
+	recorder      events.Recorder
+}
+
+// NewAddOnFeatureDiscoveryController returns an instance of addOnFeatureDiscoveryController
+func NewAddOnFeatureDiscoveryController(
+	clusterClient clientset.Interface,
+	clusterInformer clusterv1informer.ManagedClusterInformer,
+	addOnInformers addoninformerv1alpha1.ManagedClusterAddOnInformer,
+	recorder events.Recorder,
+) factory.Controller {
+	c := &addOnFeatureDiscoveryController{
+		clusterClient: clusterClient,
+		clusterLister: clusterInformer.Lister(),
+		addOnLister:   addOnInformers.Lister(),
+		recorder:      recorder,
+	}
+
+	return factory.New().
+		WithInformersQueueKeyFunc(
+			func(obj runtime.Object) string {
+				accessor, _ := meta.Accessor(obj)
+				return accessor.GetName()
+			},
+			clusterInformer.Informer()).
+		WithInformersQueueKeyFunc(
+			func(obj runtime.Object) string {
+				key, _ := cache.MetaNamespaceKeyFunc(obj)
+				return key
+			},
+			addOnInformers.Informer()).
+		WithSync(c.sync).
+		ResyncEvery(10*time.Minute).
+		ToController("AddOnFeatureDiscoveryController", recorder)
+}
+
+func (c *addOnFeatureDiscoveryController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	// The value of queueKey might be
+	// 1) equal to the default queuekey. It is triggered by resync every 10 minutes;
+	// 2) in format: namespace/name. It indicates the event source is a ManagedClusterAddOn;
+	// 3) in format: name. It indicates the event source is a ManagedCluster;
+	queueKey := syncCtx.QueueKey()
+	namespace, name, err := cache.SplitMetaNamespaceKey(queueKey)
+	if err != nil {
+		utilruntime.HandleError(err)
+		return nil
+	}
+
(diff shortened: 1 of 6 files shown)


pkg/spoke/addon/discovery_controller.go
```diff
@@ -0,0 +1,202 @@
...
+		// addon is deleting
+		if !addOn.DeletionTimestamp.IsZero() {
+			continue
+		}
+		key := addOnFeaturePrefix + addOn.Name
+		newLabels[key] = getAddOnLabelValue(addOn)
+	}
+
+	// no work if the cluster labels have no change
+	if (len(newLabels) == 0 && len(cluster.Labels) == 0) || reflect.DeepEqual(newLabels, cluster.Labels) {
```
qiujian16: do we need to compare length here?

pkg/spoke/addon/discovery_controller.go
```diff
@@ -0,0 +1,202 @@
...
+
+	return factory.New().
+		WithInformers(hubClusterInformer.Informer()).
+		WithInformersQueueKeyFunc(
+			func(obj runtime.Object) string {
+				accessor, _ := meta.Accessor(obj)
+				return accessor.GetName()
+			},
+			hubAddOnInformers.Informer()).
+		WithSync(c.sync).
```
qiujian16: I think you would need a resync here to handle unexpected deletion?

pkg/spoke/addon/discovery_controller.go
```diff
@@ -0,0 +1,202 @@
...
+
+func getAddOnLabelValue(addOn *addonv1alpha1.ManagedClusterAddOn) string {
+	availableCondition := meta.FindStatusCondition(addOn.Status.Conditions, addonv1alpha1.ManagedClusterAddOnConditionAvailable)
+	if availableCondition == nil {
+		return "unreachable"
+	}
+
+	switch availableCondition.Status {
+	case metav1.ConditionTrue:
+		return "available"
```
qiujian16: make it const

pkg/hub/addon/discovery_controller.go
```diff
@@ -0,0 +1,213 @@
...
+	c := &addOnFeatureDiscoveryController{
+		clusterClient: clusterClient,
+		clusterLister: clusterInformer.Lister(),
+		addOnLister:   addOnInformers.Lister(),
+		recorder:      recorder,
+	}
+
+	return factory.New().
+		WithInformers(clusterInformer.Informer()).
+		WithInformersQueueKeyFunc(
```
qiujian16, line 60: two cluster informers?

pkg/hub/addon/discovery_controller.go
```diff
@@ -0,0 +1,213 @@
...
+
+func (c *addOnFeatureDiscoveryController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	queueKey := syncCtx.QueueKey()
+	namespace, name, err := cache.SplitMetaNamespaceKey(queueKey)
+	if err != nil {
+		utilruntime.HandleError(err)
+		return nil
+	}
+
+	if len(namespace) > 0 {
```
qiujian16: it needs doc here, otherwise it is really strange

pkg/hub/addon/discovery_controller.go
```diff
@@ -0,0 +1,213 @@
...
+		// sync a particular addon
+		return c.syncAddOn(ctx, namespace, name)
+	}
+
+	// sync all labels of the managed cluster
+	clusterName := name
```
qiujian16: this seems redundant

pkg/hub/addon/discovery_controller.go
```diff
@@ -0,0 +1,213 @@
...
+		// addon is deleted
+		return c.deleteAddOnLabel(ctx, clusterName, addOnName)
+	}
+	if err != nil {
+		return err
+	}
+
+	// addon is deleting
+	if !addOn.DeletionTimestamp.IsZero() {
+		return c.deleteAddOnLabel(ctx, clusterName, addOnName)
```
qiujian16: it seems to me that using https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourcemerge/object_merger.go#L138 can be much easier.

pkg/spoke/addon/discovery_controller.go
elgnay: Yes. An empty map is not equal to nil, that will result in unnecessary update.

pkg/spoke/addon/discovery_controller.go
elgnay: Added.

pkg/spoke/addon/discovery_controller.go
elgnay: Done.

pkg/hub/addon/discovery_controller.go
elgnay: Removed the redundant one.

pkg/hub/addon/discovery_controller.go
elgnay: Done.

pkg/hub/addon/discovery_controller.go
elgnay: Removed.

pkg/hub/addon/discovery_controller.go
elgnay: Code updated.

pkg/hub/addon/discovery_controller.go
```diff
@@ -0,0 +1,246 @@
...
+		// addon is deleting
+		if !addOn.DeletionTimestamp.IsZero() {
+			continue
+		}
+		key := addOnFeaturePrefix + addOn.Name
+		newLabels[key] = getAddOnLabelValue(addOn)
+	}
+
+	// no work if the cluster labels have no change
+	if (len(newLabels) == 0 && len(cluster.Labels) == 0) || reflect.DeepEqual(newLabels, cluster.Labels) {
```
qiujian16: can we use MergeMap here also?

pkg/hub/addon/discovery_controller.go
```diff
@@ -0,0 +1,246 @@
...
+		syncCtx.Queue().Add(cluster.Name)
+	}
+	return nil
+}
+
+func (c *addOnFeatureDiscoveryController) syncAddOn(ctx context.Context, clusterName, addOnName string) error {
+	klog.V(4).Infof("Reconciling addOn %q", addOnName)
+
+	addOn, err := c.addOnLister.ManagedClusterAddOns(clusterName).Get(addOnName)
+	if errors.IsNotFound(err) {
```
qiujian16: I might get cluster here, merge labels, then update cluster. So we do not need to differentiate deleteAddonLabel and updateAddonLabel.

pkg/hub/addon/discovery_controller.go
elgnay: Yes. The code updated.

pkg/hub/addon/discovery_controller.go
elgnay: Code updated as suggested.

pkg/hub/addon/discovery_controller.go
qiujian16: nit: if you  use apiequality.sementic.DeepEqual you would not need to compare the length

pkg/hub/addon/discovery_controller.go
qiujian16: what i  meant is

```go

labels := map[string]string{}
addOn, err := c.addOnLister.ManagedClusterAddOns(clusterName).Get(addOnName)
switch {
case errors.IsNotFound(err):
  labels[addonName + "-"] = ""
case err != nil:
  return err
case !addOn.DeletionTimestamp.IsZero():
  labels[addonName + "-"] = ""
default:
  labels[addonName] = getAddOnLabelValue(addOn)
}

cluster, err := c.clusterLister.Get(clusterName)
if err != nil {
	return fmt.Errorf("unable to find cluster with name %q: %w", clusterName, err)
}

resourcemerge.MergeMap(&modified, &cluster.Labels,  labels)

if modified {
 // update things
}

```

pkg/hub/addon/discovery_controller.go
qiujian16: I might do like:

```go
newLabels := map[string]string{}
addOns, err := c.addOnLister.ManagedClusterAddOns(clusterName).List(labels.Everything())
	if err != nil {
		return fmt.Errorf("unable to list addOns of cluster %q: %w", clusterName, err)
	}
	for _, addOn := range addOns {
		// addon is deleting
		if !addOn.DeletionTimestamp.IsZero() {
			continue
		}
		key := addOnFeaturePrefix + addOn.Name
		newLabels[key] = getAddOnLabelValue(addOn)
	}

newCluster := cluster.DeepCopy()
for key, value := range newCluster.Labels {
		if !strings.HasPrefix(key, addOnFeaturePrefix) {
			continue
		}
              if _, ok := newLabels[key]; !ok {
                    newLabels[key + "-"] = value
              }
		
	}

resourcemerge.MergeMap(&modified, &newCluster.Labels,  newLabels)

if modified {
// update things
}

```

qiujian16 (commented):
/approve
/lgtm

PR #172, opened 2021-10-12
re-create addon client cert once bootstrap kubeconfig changes

elgnay:

### DIFF
--- a/pkg/clientcert/cert_controller.go
+++ b/pkg/clientcert/cert_controller.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509/pkix"
 	"fmt"
 	"math/rand"
+	"reflect"
 	"time"
 
 	"github.com/openshift/library-go/pkg/controller/factory"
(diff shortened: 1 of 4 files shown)


pkg/clientcert/cert_controller.go
```diff
@@ -187,7 +191,12 @@ func (c *clientCertificateController) sync(ctx context.Context, syncCtx factory.
 	// create a csr to request new client certificate if
 	// a. there is no valid client certificate issued for the current cluster/agent
 	// b. client certificate exists and has less than a random percentage range from 20% to 25% of its life remaining
-	if c.hasValidClientCertificate(secret) {
+
+	hasValidClientCert := c.hasValidClientCertificate(secret)
+	switch {
+	case hasValidClientCert && c.AdditonalSecretDataSensitive && !c.hasAdditonalSecretData(secret):
```
qiujian16: maybe

```go
switch{
case !hasValidClientCert:
    // record event and break
case c.AdditonalSecretDataSensitive && !c.hasAdditonalSecretData(secret):
   // record event and break
default:
  ...
}
```

elgnay: Updated.

PR #174, opened 2021-10-25
exclude the cordoned nodes when calc node allocatable

skeeey:


skeeey:
refer to https://github.com/open-cluster-management-io/community/issues/77

pkg/spoke/managedcluster/node_controller.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (c *managedClusterNodeController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	managedCluster, err := c.hubClusterLister.Get(c.clusterName)
+	if err != nil {
+		return fmt.Errorf("unable to get managed cluster with name %q from hub: %w", c.clusterName, err)
+	}
+
+	// current managed cluster is not accepted, do nothing.
+	if !meta.IsStatusConditionTrue(managedCluster.Status.Conditions, clusterv1.ManagedClusterConditionHubAccepted) {
```
qiujian16: should this be Available?

qiujian16 (commented):
why do we separate this  to another controller? It will double the update call, right?

pkg/spoke/managedcluster/node_controller.go
skeeey: yes, the `Available` is more suitable than `Accepted`

skeeey:
> why do we separate this to another controller? It will double the update call, right?

I think separate this make the code logic more clear, currently, we maintain this in the cluster joining controller, but collecting the node resource and handling cluster join event may be two irrelevant tasks

when the cluster starts to register, this will add an update call, but if we use  cluster `Available` as the judgement condition, I think we can stagger this update from others

what's your opinion? do you think we should keep this in the joining controller?

qiujian16:
I think we can move it as part of health check (available) controller

skeeey:
agree with you, and I think we can also put the get version in health check controller, thus the joining controller will only handle the cluster joining

PR #182, opened 2021-12-02
deploy control plane and agent in two clusters

zhujian7:
This PR supports deploying registration control plane and agent in two separate clusters and in different namespaces. It is good for new guys to understand the registration.

cc @qiujian16 @skeeey

### DIFF
--- a/deploy/spoke/clusterrole_binding.yaml
+++ b/deploy/spoke/clusterrole_binding.yaml
@@ -10,4 +10,3 @@ subjects:
   - kind: ServiceAccount
     name: spoke-agent-sa
     namespace: open-cluster-management-agent
-  
--- a/deploy/spoke/clusterrole_binding.yaml
+++ b/deploy/spoke/clusterrole_binding.yaml
@@ -9,5 +9,5 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: spoke-agent-sa
-    namespace: open-cluster-management
+    namespace: open-cluster-management-agent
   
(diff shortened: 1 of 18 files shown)


zhujian7:
> 	// defaultSpokeComponentNamespace is the default namespace in which the spoke agent is deployed
	defaultSpokeComponentNamespace = "open-cluster-management"

I am not quite sure if https://github.com/open-cluster-management-io/registration/blob/main/pkg/spoke/spokeagent.go#L45 should be changed to "open-cluster-management-agent"
Is this `defaultSpokeComponentNamespace` only useful for tests?

skeeey:
> > // defaultSpokeComponentNamespace is the default namespace in which the spoke agent is deployed
> > defaultSpokeComponentNamespace = "open-cluster-management"
> 
> I am not quite sure if https://github.com/open-cluster-management-io/registration/blob/main/pkg/spoke/spokeagent.go#L45 should be changed to "open-cluster-management-agent" Is this `defaultSpokeComponentNamespace` only useful for tests?

We also use `open-cluster-management-agent` as the default namespace in registration-operator, so I think we may change this to keep the consistent

/cc @qiujian16 what's your opinion?

deploy/spoke/clusterrole_binding.yaml
```diff
@@ -9,5 +9,5 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: spoke-agent-sa
-    namespace: open-cluster-management
+    namespace: open-cluster-management-agent
```
skeeey: need a new line

qiujian16:
yes, I think we should change it to open-cluster-management-agent

PR #186, opened 2021-12-22
enhance webhook to set the value of taint timeAdded

elgnay:


pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -58,18 +67,115 @@ func (a *ManagedClusterMutatingAdmissionHook) Admit(req *admissionv1beta1.Admiss
...
+		}
+	}
+
+	var jsonPatches []jsonPatchOperation
+	now := metav1.NewTime(nowFunc())
+	for index, taint := range managedCluster.Spec.Taints {
+		// reuse the existing value of timeAdded if all the key/value/effect are matched
+		// otherwise reset the value
+		timeAdded := getTaintTimeAdded(oldManagedCluster, taint.Key, taint.Value, taint.Effect, now)
+		if reflect.DeepEqual(timeAdded, taint.TimeAdded) {
```
qiujian16: use apiequality

pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -58,18 +67,115 @@ func (a *ManagedClusterMutatingAdmissionHook) Admit(req *admissionv1beta1.Admiss
...
 	}
 
+	var jsonPatches []jsonPatchOperation
+
 	// If LeaseDurationSeconds value is zero, update it to 60 by default
 	if managedCluster.Spec.LeaseDurationSeconds == 0 {
-		status.Patch = []byte(defaultLeaseDurationSecondsPatch)
-		pt := admissionv1beta1.PatchTypeJSONPatch
-		status.PatchType = &pt
+		jsonPatches = append(jsonPatches, newLeaseDurationSecondsJsonPatch(60))
```
qiujian16: this can be removed this we already set default value

pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -58,18 +67,115 @@ func (a *ManagedClusterMutatingAdmissionHook) Admit(req *admissionv1beta1.Admiss
...
+			return nil, err
+		}
+	}
+
+	var jsonPatches []jsonPatchOperation
+	now := metav1.NewTime(nowFunc())
+	for index, taint := range managedCluster.Spec.Taints {
+		// reuse the existing value of timeAdded if all the key/value/effect are matched
+		// otherwise reset the value
+		timeAdded := getTaintTimeAdded(oldManagedCluster, taint.Key, taint.Value, taint.Effect, now)
```
qiujian16: I think we just need to set timeAdded if it is nil? why do we need to compare it with the old value. To make it immutable?

pkg/webhook/cluster/mutating_webhook.go
elgnay: Done.

pkg/webhook/cluster/mutating_webhook.go
elgnay: Removed.

pkg/webhook/cluster/mutating_webhook.go
elgnay: Yes. After a taint is created, the timeAdded should be immutable unless its key/value/effect changes. It should be maintained by the webhook rather than user.

qiujian16: i think we need validatingwebhook to do the immutable check rather than implicitly mutate it

elgnay: Validation has been added in the mutating webhook instead of validatingwebhook.

pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -58,18 +67,122 @@ func (a *ManagedClusterMutatingAdmissionHook) Admit(req *admissionv1beta1.Admiss
...
+	return nil, status
+}
+
 // Initialize is called by generic-admission-server on startup to setup initialization that managedclusters webhook needs.
 func (a *ManagedClusterMutatingAdmissionHook) Initialize(kubeClientConfig *rest.Config, stopCh <-chan struct{}) error {
 	// do nothing
 	return nil
 }
+
+func findTaintWithKey(managedCluster *clusterv1.ManagedCluster, key string) *clusterv1.Taint {
```
qiujian16: put this as a common helper func. I feel you will also need it in controller's code

qiujian16: and the func is better to be `FindTaintByKey(taints, key)`

elgnay: Done.

qiujian16: Did you push the commit? I do not see the change here yet.

elgnay: The change is in place now.

pkg/helpers/helpers.go
```diff
@@ -321,3 +321,17 @@ func ManagedClusterAssetFn(fs embed.FS, managedClusterName string) resourceapply
 		return assets.MustCreateAssetFromTemplate(name, template, config).Data, nil
 	}
 }
+
+// FindTaintByKey returns a taint if the managed cluster has a taint with the given key.
+func FindTaintByKey(managedCluster *clusterv1.ManagedCluster, key string) *clusterv1.Taint {
```
qiujian16, line 326: nit: I think  `func FindTaintByKey(taints []clusterv1.Taint, key string) *clusterv1.Taint` is more natural

PR #188, opened 2022-01-05
add taints controller

JiahaoWei-RH:
Signed-off-by: JiaHaoWei 1335649406@qq.com
Implement the Taints mentioned in  #https://github.com/open-cluster-management-io/community/issues/48, and add ut and integration tests

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -246,3 +246,47 @@ func FindTaintByKey(managedCluster *clusterv1.ManagedCluster, key string) *clust
 	}
 	return nil
 }
+
+func IsTaintEqual(taint1, taint2 clusterv1.Taint) bool {
+	// Ignore the comparison of time
+	return taint1.Key == taint2.Key && taint1.Value == taint2.Value && taint1.Effect == taint2.Effect
+}
+
+// AddTaints add taints to the specified slice, if it did not already exist.
+// Return a boolean indicating whether the slice has been updated.
+func AddTaints(taints *[]clusterv1.Taint, taint clusterv1.Taint) bool {
+	if taints == nil || *taints == nil {
+		*taints = make([]clusterv1.Taint, 0)
+	}
+	if FindTaint(*taints, taint) != nil {
+		return false
+	}
+	*taints = append(*taints, taint)
+	return true
+}
+
+func RemoveTaints(taints *[]clusterv1.Taint, targets ...clusterv1.Taint) (updated bool) {
+	if taints == nil || len(*taints) == 0 || len(targets) == 0 {
+		return false
+	}
+
+	newTaints := make([]clusterv1.Taint, 0)
+	for _, v := range *taints {
+		if FindTaint(targets, v) == nil {
+			newTaints = append(newTaints, v)
+		}
+	}
+	updated = len(*taints) != len(newTaints)
+	*taints = newTaints
+	return updated
+}
+
+func FindTaint(taints []clusterv1.Taint, taint clusterv1.Taint) *clusterv1.Taint {
+	for i := range taints {
+		if IsTaintEqual(taints[i], taint) {
+			return &taints[i]
+		}
+	}
+
+	return nil
+}
--- a/pkg/helpers/testing/testinghelpers.go
+++ b/pkg/helpers/testing/testinghelpers.go
@@ -90,14 +90,27 @@ func NewAcceptedManagedCluster() *clusterv1.ManagedCluster {
 
 func NewAvailableManagedCluster() *clusterv1.ManagedCluster {
 	managedCluster := NewAcceptedManagedCluster()
-	availableCondtion := NewManagedClusterCondition(
+	availableCondition := NewManagedClusterCondition(
 		clusterv1.ManagedClusterConditionAvailable,
 		"True",
 		"ManagedClusterAvailable",
 		"Managed cluster is available",
 		nil,
 	)
-	managedCluster.Status.Conditions = append(managedCluster.Status.Conditions, availableCondtion)
+	managedCluster.Status.Conditions = append(managedCluster.Status.Conditions, availableCondition)
+	return managedCluster
+}
+
+func NewUnAvailableManagedCluster() *clusterv1.ManagedCluster {
+	managedCluster := NewAcceptedManagedCluster()
+	condition := NewManagedClusterCondition(
+		clusterv1.ManagedClusterConditionAvailable,
+		"False",
+		"ManagedClusterUnAvailable",
+		"Managed cluster is Unavailable",
+		nil,
+	)
+	managedCluster.Status.Conditions = append(managedCluster.Status.Conditions, condition)
 	return managedCluster
 }
 
--- a/pkg/hub/taint/controller.go
+++ b/pkg/hub/taint/controller.go
@@ -1,0 +1,96 @@
+package taint
+
+import (
+	"context"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	clientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	informerv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	listerv1 "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	v1 "open-cluster-management.io/api/cluster/v1"
+	"open-cluster-management.io/registration/pkg/helpers"
+)
+
+var (
+	UnavailableTaint = v1.Taint{
+		Key:    v1.ManagedClusterTaintUnavailable,
+		Effect: v1.TaintEffectNoSelect,
+	}
+
+	UnreachableTaint = v1.Taint{
+		Key:    v1.ManagedClusterTaintUnreachable,
+		Effect: v1.TaintEffectNoSelect,
+	}
+)
+
+// taintController
+type taintController struct {
+	clusterClient clientset.Interface
+	clusterLister listerv1.ManagedClusterLister
+	eventRecorder events.Recorder
+}
+
+// NewTaintController creates a new taint controller
+func NewTaintController(
+	clusterClient clientset.Interface,
+	clusterInformer informerv1.ManagedClusterInformer,
+	recorder events.Recorder) factory.Controller {
+	c := &taintController{
+		clusterClient: clusterClient,
+		clusterLister: clusterInformer.Lister(),
+		eventRecorder: recorder.WithComponentSuffix("taint-controller"),
+	}
+	return factory.New().
+		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
+			accessor, _ := meta.Accessor(obj)
+			return accessor.GetName()
+		}, clusterInformer.Informer()).
+		WithSync(c.sync).
+		ToController("taintController", recorder)
+}
+
+func (c *taintController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	managedClusterName := syncCtx.QueueKey()
+	klog.V(4).Infof("Reconciling ManagedCluster %s", managedClusterName)
+	managedCluster, err := c.clusterLister.Get(managedClusterName)
+	if errors.IsNotFound(err) {
+		// Spoke cluster not found, could have been deleted, do nothing.
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	if !managedCluster.DeletionTimestamp.IsZero() {
+		return nil
+	}
+
+	managedCluster = managedCluster.DeepCopy()
+	newTaints := managedCluster.Spec.Taints
+	cond := meta.FindStatusCondition(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable)
+	var updated bool
+
+	switch {
+	case cond == nil || cond.Status == metav1.ConditionUnknown:
+		updated = helpers.RemoveTaints(&newTaints, UnavailableTaint)
+		updated = helpers.AddTaints(&newTaints, UnreachableTaint) || updated
+	case cond.Status == metav1.ConditionFalse:
+		updated = helpers.RemoveTaints(&newTaints, UnreachableTaint)
+		updated = helpers.AddTaints(&newTaints, UnavailableTaint) || updated
+	case cond.Status == metav1.ConditionTrue:
+		updated = helpers.RemoveTaints(&newTaints, UnavailableTaint, UnreachableTaint)
+	}
+
+	if updated {
+		managedCluster.Spec.Taints = newTaints
+		if _, err = c.clusterClient.ClusterV1().ManagedClusters().Update(ctx, managedCluster, metav1.UpdateOptions{}); err != nil {
+			return err
+		}
+		c.eventRecorder.Eventf("ManagedClusterConditionAvailableUpdated", "Update the original taints to the %+v", newTaints)
+	}
+	return nil
+}
--- a/test/integration/taint_add_test.go
+++ b/test/integration/taint_add_test.go
@@ -1,0 +1,148 @@
+package integration_test
+
+import (
+	"context"
+	"fmt"
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+	v1 "open-cluster-management.io/api/cluster/v1"
+	"open-cluster-management.io/registration/pkg/helpers"
+	"open-cluster-management.io/registration/pkg/hub/taint"
+	"open-cluster-management.io/registration/pkg/spoke"
+	"open-cluster-management.io/registration/test/integration/util"
+	"path"
+	"time"
+)
+
+var _ = ginkgo.Describe("ManagedCluster Taints Update", func() {
+	var managedClusterName string
+	var hubKubeconfigSecret string
+	var hubKubeconfigDir string
+
+	ginkgo.BeforeEach(func() {
+		managedClusterName = fmt.Sprintf("managedcluster-%s", rand.String(6))
+		hubKubeconfigSecret = fmt.Sprintf("%s-secret", managedClusterName)
+		hubKubeconfigDir = path.Join(util.TestDir, "leasetest", fmt.Sprintf("%s-config", managedClusterName))
+	})
+
+	ginkgo.It("ManagedCluster taint should be updated automatically", func() {
+		ctx, stop := context.WithCancel(context.Background())
+		// run registration agent
+		go func() {
+			agentOptions := spoke.SpokeAgentOptions{
+				ClusterName:              managedClusterName,
+				BootstrapKubeconfig:      bootstrapKubeConfigFile,
+				HubKubeconfigSecret:      hubKubeconfigSecret,
+				HubKubeconfigDir:         hubKubeconfigDir,
+				ClusterHealthCheckPeriod: 1 * time.Minute,
+			}
+			err := agentOptions.RunSpokeAgent(ctx, &controllercmd.ControllerContext{
+				KubeConfig:    spokeCfg,
+				EventRecorder: util.NewIntegrationTestEventRecorder("cluster-tainttest"),
+			})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		}()
+
+		gomega.Eventually(func() error {
+			spokeCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
+			if err != nil {
+				return err
+			}
+			spokeCluster.Spec.HubAcceptsClient = true
+			_, err = clusterClient.ClusterV1().ManagedClusters().Update(context.TODO(), spokeCluster, metav1.UpdateOptions{})
+			return err
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.Succeed())
+
+		gomega.Eventually(func() error {
+			managedCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
+			if err != nil {
+				return err
+			}
+			if len(managedCluster.Spec.Taints) != 1 {
+				return fmt.Errorf("managedCluster taints len is not 1")
+			}
+			if !helpers.IsTaintEqual(managedCluster.Spec.Taints[0], taint.UnreachableTaint) {
+				return fmt.Errorf("the %+v is not equal to UnreachableTaint", managedCluster.Spec.Taints[0])
+			}
+			return nil
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeNil())
+
+		gomega.Eventually(func() error {
+			if err := authn.ApproveSpokeClusterCSR(kubeClient, managedClusterName, time.Hour*24); err != nil {
+				return err
+			}
+			return nil
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeNil())
+
+		// The managed cluster is available, so taint is expected to be empty
+		gomega.Eventually(func() bool {
+			managedCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
+			if err != nil {
+				return false
+			}
+			if len(managedCluster.Spec.Taints) != 0 {
+				return false
+			}
+			return true
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeTrue())
+
+		stop()
+
+		gomega.Eventually(func() error {
+			managedCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
+			if err != nil {
+				return err
+			}
+			condition := metav1.Condition{
+				Type:   v1.ManagedClusterConditionAvailable,
+				Status: metav1.ConditionFalse,
+				Reason: "ForTest",
+			}
+			meta.SetStatusCondition(&(managedCluster.Status.Conditions), condition)
+			_, err = clusterClient.ClusterV1().ManagedClusters().UpdateStatus(context.Background(), managedCluster, metav1.UpdateOptions{})
+			return err
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeNil())
+
+		gomega.Eventually(func() error {
+			managedCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
+			if err != nil {
+				return err
+			}
+			if len(managedCluster.Spec.Taints) != 1 {
+				return fmt.Errorf("managedCluster taints len is not 1")
+			}
+			if !helpers.IsTaintEqual(managedCluster.Spec.Taints[0], taint.UnavailableTaint) {
+				return fmt.Errorf("the %+v is not equal to UnavailableTaint", managedCluster.Spec.Taints[0])
+			}
+			return nil
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeNil())
+
+		gomega.Eventually(func() error {
+			spokeCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
+			if err != nil {
+				return err
+			}
+			spokeCluster.Spec.LeaseDurationSeconds = 1
+			_, err = clusterClient.ClusterV1().ManagedClusters().Update(context.TODO(), spokeCluster, metav1.UpdateOptions{})
+			return err
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.Succeed())
+
+		gomega.Eventually(func() error {
+			managedCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
+			if err != nil {
+				return err
+			}
+			if len(managedCluster.Spec.Taints) != 1 {
+				return fmt.Errorf("managedCluster taints len is not 1")
+			}
+			if !helpers.IsTaintEqual(managedCluster.Spec.Taints[0], taint.UnreachableTaint) {
+				return fmt.Errorf("the %+v is not equal to UnreachableTaint", managedCluster.Spec.Taints[0])
+			}
+			return nil
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeNil())
+	})
+})
--- a/pkg/helpers/helpers_test.go
+++ b/pkg/helpers/helpers_test.go
@@ -383,3 +383,146 @@ func getApplyFileNames(applyFiles map[string]runtime.Object) []string {
 	}
 	return keys
 }
+
+var (
+	UnavailableTaint = clusterv1.Taint{
+		Key:    clusterv1.ManagedClusterTaintUnavailable,
+		Effect: clusterv1.TaintEffectNoSelect,
+	}
+
+	UnreachableTaint = clusterv1.Taint{
+		Key:    clusterv1.ManagedClusterTaintUnreachable,
+		Effect: clusterv1.TaintEffectNoSelect,
+	}
+)
+
+func TestIsTaintsEqual(t *testing.T) {
+	cases := []struct {
+		name    string
+		taints1 []clusterv1.Taint
+		taints2 []clusterv1.Taint
+		expect  bool
+	}{
+		{
+			name:    "two empty taints",
+			taints1: []clusterv1.Taint{},
+			taints2: []clusterv1.Taint{},
+			expect:  true,
+		},
+		{
+			name:    "two nil taints",
+			taints1: nil,
+			taints2: nil,
+			expect:  true,
+		},
+		{
+			name:    "len(taints1) = 1, len(taints2) = 0",
+			taints1: []clusterv1.Taint{UnavailableTaint},
+			taints2: []clusterv1.Taint{},
+			expect:  false,
+		},
+		{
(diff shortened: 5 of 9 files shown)


pkg/hub/taints/doc.go
```diff
@@ -0,0 +1 @@
+package taints
```
zhiweiyin318: why need this？

pkg/helpers/testing/testinghelpers.go
```diff
@@ -90,14 +90,27 @@ func NewAcceptedManagedCluster() *clusterv1.ManagedCluster {
...
 		nil,
 	)
-	managedCluster.Status.Conditions = append(managedCluster.Status.Conditions, availableCondtion)
+	managedCluster.Status.Conditions = append(managedCluster.Status.Conditions, availableCondition)
+	return managedCluster
+}
+
+func NewUnAvailableManagedCluster() *clusterv1.ManagedCluster {
+	managedCluster := NewAcceptedManagedCluster()
+	availableCondition := NewManagedClusterCondition(
```
zhiweiyin318: unavailableCondition or condition? availableCondition makes some confusion.

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	"k8s.io/klog/v2"
+	clientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	informerv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	listerv1 "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	v1 "open-cluster-management.io/api/cluster/v1"
+	"time"
+)
+
+const (
+	NoCondition      = "NoManagedClusterConditionAvailable"
```
zhiweiyin318: these const vars are taint value, right?  it's better to add some comments/doc here.

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+			newTaints = append(newTaints, t1)
+		}
+	}
+	if c.isTaintsEqual(managedCluster.Spec.Taints, newTaints) {
+		return nil
+	}
+
+	managedCluster.Spec.Taints = newTaints
+	managedCluster, err = c.clusterClient.ClusterV1().ManagedClusters().Update(ctx, managedCluster, metav1.UpdateOptions{})
```
zhiweiyin318: managedCluster is not used, maybe `_ , err = c.clusterClient.ClusterV1().ManagedClusters().Update(ctx, managedCluster, metav1.UpdateOptions{})` is better ?

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		if v.Key == isExistKey {
+			exist = true
+		}
+	}
+
+	return ans, exist
+}
+
+func (c *taintsController) isIsStatusConditionUnknown(conditions []metav1.Condition, conditionType string) bool {
```
zhiweiyin318: isStatusConditionUnknown?

pkg/hub/taints/doc.go
JiahaoWei-RH: Looks like it should be deleted, because no description was added.

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+
+const (
+	NoCondition      = "NoManagedClusterConditionAvailable"
+	ConditionUnknown = "ManagedClusterConditionAvailableUnknown"
+	ConditionFalse   = "ManagedClusterConditionAvailableFalse"
+)
+
+// taintsController
+type taintsController struct {
+	kubeClient    kubernetes.Interface
```
skeeey: we don't need this client

pkg/hub/taints/controller.go
JiahaoWei-RH: OK, I will add later.

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	if c.isTaintsEqual(managedCluster.Spec.Taints, newTaints) {
+		return nil
+	}
+
+	managedCluster.Spec.Taints = newTaints
+	managedCluster, err = c.clusterClient.ClusterV1().ManagedClusters().Update(ctx, managedCluster, metav1.UpdateOptions{})
+	return err
+}
+
+func (c *taintsController) deleteTaintAndJudgeExist(taints []v1.Taint, isExistKey string, deleteKeys ...string) ([]v1.Taint, bool) {
```
skeeey, line 118: would you split this method into two methods? one is used to delete and the other is used to judge exist, I think it will be more clear

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if err != nil {
+		return err
+	}
+	managedCluster = managedCluster.DeepCopy()
+	newTaints := managedCluster.Spec.Taints
+
+	switch {
+	case meta.IsStatusConditionTrue(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable):
+		newTaints, _ = c.deleteTaintAndJudgeExist(newTaints, "",
```
skeeey: would you split this method into two methods? one is used to delete and the other is used to judge exist, I think it will be more clear,
e.g. here will be `c.deleteTaints(newTaints , v1.ManagedClusterTaintUnreachable, v1.ManagedClusterTaintUnavailable)`

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+			v1.ManagedClusterTaintUnreachable, v1.ManagedClusterTaintUnavailable)
+	case meta.IsStatusConditionFalse(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable):
+		var exist bool
+		newTaints, exist = c.deleteTaintAndJudgeExist(newTaints,
+			v1.ManagedClusterTaintUnavailable, v1.ManagedClusterTaintUnreachable)
```
skeeey: here will be 
```
c.deleteTaints(newTaints, v1.ManagedClusterTaintUnavailable)
if !hasTaints(newTaints, v1.ManagedClusterTaintUnavailable) {
  // ...
}
```

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+				Key:    v1.ManagedClusterTaintUnavailable,
+				Value:  ConditionFalse,
+				Effect: v1.TaintEffectNoSelect,
+				TimeAdded: metav1.Time{
+					Time: time.Now(),
+				},
+			}
+			newTaints = append(newTaints, t1)
+		}
+	case c.isIsStatusConditionUnknown(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable) ||
```
skeeey: we may not need to wrap the `meta.IsStatusConditionPresentAndEqual`

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+		meta.FindStatusCondition(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable) == nil:
```
skeeey: this may need another case to switch

skeeey: and do we really need to judge this?

pkg/hub/taints/controller.go
skeeey: and if update success, you may use `eventRecorder` to recorder this

pkg/hub/taints/doc.go
skeeey: we usually add comments here to explain the function of this package

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+				Value:  value,
+				Effect: v1.TaintEffectNoSelect,
+				TimeAdded: metav1.Time{
+					Time: time.Now(),
+				},
+			}
+			newTaints = append(newTaints, t1)
+		}
+	}
+	if c.isTaintsEqual(managedCluster.Spec.Taints, newTaints) {
```
skeeey: how about use `equality.Semantic.DeepEqual`

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+			accessor, _ := meta.Accessor(obj)
+			return accessor.GetName()
+		}, clusterInformer.Informer()).
+		WithSync(c.sync).
+		ToController("taintsController", recorder)
+}
+
+func (c *taintsController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	managedClusterName := syncCtx.QueueKey()
+	klog.V(4).Infof("Reconciling Taints %s", managedClusterName)
```
elgnay: "Reconciling Taints %s" -> "Reconciling ManagedCluster %s"?

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	case meta.IsStatusConditionFalse(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable):
+		var exist bool
+		newTaints, exist = c.deleteTaintAndJudgeExist(newTaints,
+			v1.ManagedClusterTaintUnavailable, v1.ManagedClusterTaintUnreachable)
+		if !exist {
+			t1 := v1.Taint{
+				Key:    v1.ManagedClusterTaintUnavailable,
+				Value:  ConditionFalse,
+				Effect: v1.TaintEffectNoSelect,
+				TimeAdded: metav1.Time{
```
elgnay: Do NOT set TimeAdded. Instead it will be maintained by a webhook.

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+		value := ConditionUnknown
+		if meta.FindStatusCondition(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable) == nil {
+			value = NoCondition
+		}
+		if !exist {
+			t1 := v1.Taint{
+				Key:    v1.ManagedClusterTaintUnreachable,
+				Value:  value,
+				Effect: v1.TaintEffectNoSelect,
+				TimeAdded: metav1.Time{
```
elgnay: Do NOT set TimeAdded. Instead it will be maintained by a webhook.

pkg/hub/taints/controller.go
elgnay: When comparing two taints, the TimeAdded should be ignored since it will be set by a webhook.

test/integration/taint_add_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+		gomega.Eventually(func() bool {
+			managedCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
+			if err != nil {
+				return false
+			}
+			if len(managedCluster.Spec.Taints) != 1 {
+				return false
+			}
+			if managedCluster.Spec.Taints[0] != (v1.Taint{
```
elgnay: Do not compare two objects in this way. Just compare their fields, including Key, Value and Effect, one by one.

test/integration/taint_add_test.go
```diff
@@ -0,0 +1,167 @@
...
+				Key:       "cluster.open-cluster-management.io/unreachable",
```
elgnay: Please reference the pre-defined taint in the cluster api

test/integration/taint_add_test.go
```diff
@@ -0,0 +1,167 @@
...
+			managedCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			condition := meta.FindStatusCondition(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable)
+			gomega.Expect(condition).ShouldNot(gomega.BeNil())
+			gomega.Expect(condition.Status).Should(gomega.Equal(metav1.ConditionTrue))
+			if condition != nil {
+				condition.Status = "False"
+			}
+			meta.SetStatusCondition(&(managedCluster.Status.Conditions), *condition)
+			managedCluster, err = clusterClient.ClusterV1().ManagedClusters().UpdateStatus(context.Background(), managedCluster, metav1.UpdateOptions{})
```
elgnay: Why we need to restart the agent?

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
+package taints
```
elgnay: Use `taint` instead of `taints`.

pkg/helpers/testing/testinghelpers.go
JiahaoWei-RH: ok, Change availableCondition to condition.

pkg/hub/taints/controller.go
JiahaoWei-RH: But splitting into two methods will increase the time complexity.

test/integration/taint_add_test.go
JiahaoWei-RH: During the test, I found that if we do not restart the agent, the condition status may be changed to unknown automatically.

test/integration/taint_add_test.go
JiahaoWei-RH: OK

pkg/hub/taints/controller.go
JiahaoWei-RH: It can be split into two cases, but the logic of the two cases is very similar, only the value is different.

pkg/hub/taints/controller.go
JiahaoWei-RH: I tried to delete it once, but had an error when updating.

pkg/helpers/testing/testinghelpers.go
```diff
@@ -90,14 +90,27 @@ func NewAcceptedManagedCluster() *clusterv1.ManagedCluster {
...
+		clusterv1.ManagedClusterConditionAvailable,
+		"False",
+		"ManagedClusterUnAvailable",
+		"Managed cluster is Unavailable",
+		nil,
+	)
+	managedCluster.Status.Conditions = append(managedCluster.Status.Conditions, availableCondition)
```
qiujian16: use meta condition func helper instead of using append

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	clientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	informerv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	listerv1 "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	v1 "open-cluster-management.io/api/cluster/v1"
+	"time"
+)
+
+const (
+	NoCondition      = "NoManagedClusterConditionAvailable"
+	ConditionUnknown = "ManagedClusterConditionAvailableUnknown"
```
qiujian16: what are these for?

pkg/hub/taints/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	if (taints1 == nil) != (taints2 == nil) {
+		return false
+	}
+
+	if len(taints2) == 0 && len(taints1) == 0 {
+		return true
+	}
+
+	for i, v := range taints1 {
+		if v != taints2[i] {
```
qiujian16: I think  you should compare taint without timeadded field. It is better to have a isTaintEqual helper.

qiujian16 (commented):
I think you should build a set of taint helpers at first such as
```go
SetTaint(taints, taint)
RemoveTatint(taints, taint)
```
All these can use key as the index

Then you do not need to check if a taint exist or not in each switch. When the cluster is unavailable, always set the taint. When it is available always remove the taint.

qiujian16:
haven't review the test code yet, I think we should make the controller code cleaner which some helper funcs

pkg/hub/taints/controller.go
JiahaoWei-RH: these const are used to distinguish the two situations of NoCondition and ConditionUnknown, after discussion it was unnecessary

pkg/helpers/helpers.go
```diff
@@ -321,3 +321,69 @@ func ManagedClusterAssetFn(fs embed.FS, managedClusterName string) resourceapply
...
+func HasTaints(taints []clusterv1.Taint, target clusterv1.Taint) bool {
+	for _, v := range taints {
+		if TaintEqual(v, target) {
+			return true
+		}
+	}
+	return false
+}
+
+func AddTaints(taints *[]clusterv1.Taint, taint clusterv1.Taint) {
```
elgnay: Please return a new array instead of mutating the original one.

elgnay: Also you may need to handle the case in which a taint with same key but different value/effect already exists.

pkg/helpers/helpers.go
```diff
@@ -321,3 +321,69 @@ func ManagedClusterAssetFn(fs embed.FS, managedClusterName string) resourceapply
...
+		}
+		if !flag {
+			return false
+		}
+	}
+
+	return true
+}
+
+func HasTaints(taints []clusterv1.Taint, target clusterv1.Taint) bool {
```
elgnay: HasTaints -> HasTaint

pkg/helpers/helpers.go
```diff
@@ -321,3 +321,69 @@ func ManagedClusterAssetFn(fs embed.FS, managedClusterName string) resourceapply
...
+
+func AddTaints(taints *[]clusterv1.Taint, taint clusterv1.Taint) {
+	if HasTaints(*taints, taint) {
+		return
+	}
+	*taints = append(*taints, taint)
+	return
+}
+
+func RemoveTaints(taints *[]clusterv1.Taint, targets ...clusterv1.Taint) {
```
elgnay: Return a new array instead of mutating the original one.

pkg/helpers/testing/assertion.go
```diff
@@ -108,6 +109,16 @@ func AssertFinalizers(t *testing.T, obj runtime.Object, finalizers []string) {
 	}
 }
 
+// AssertTaints asserts the given runtime object has the expected taint
+func AssertTaints(t *testing.T, actualTaints []clusterv1.Taint, expectTaints []clusterv1.Taint) {
+	if len(actualTaints) != len(expectTaints) {
+		t.Errorf("expected taint len is %d, but actual len is %d", len(expectTaints), len(actualTaints))
+	}
+	if !helpers.IsTaintsEqual(actualTaints, expectTaints) {
+		t.Errorf("expected taint %#v, but expectTaints: %#v", expectTaints, actualTaints)
```
elgnay: but actualTaints

pkg/helpers/helpers.go
skeeey: you may return a removed flag to indicate we do the remove action

pkg/hub/taint/controller.go
```diff
@@ -0,0 +1,98 @@
...
+	informerv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	listerv1 "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	v1 "open-cluster-management.io/api/cluster/v1"
+	"open-cluster-management.io/registration/pkg/helpers"
+)
+
+var (
+	UnavailableTaint = v1.Taint{
+		Key:    v1.ManagedClusterTaintUnavailable,
+		Value:  "",
```
elgnay: This line is unnecessary.

pkg/hub/taint/controller.go
```diff
@@ -0,0 +1,98 @@
...
+		Effect: v1.TaintEffectNoSelect,
+	}
+
+	UnreachableTaint = v1.Taint{
+		Key:    v1.ManagedClusterTaintUnreachable,
+		Value:  "",
```
elgnay: This line is unnecessary.

pkg/helpers/helpers.go
skeeey: And you may also return an updated flag to indicate we do the update action

pkg/hub/taint/controller.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return err
+	}
+	managedCluster = managedCluster.DeepCopy()
+	newTaints := managedCluster.Spec.Taints
+
+	switch {
+	case meta.IsStatusConditionTrue(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable):
+		helpers.RemoveTaints(&newTaints, UnavailableTaint, UnreachableTaint)
+		c.eventRecorder.Eventf("ManagedClusterConditionAvailable conditionStatus is True", "remove UnavailableTaint and UnreachableTaint")
```
skeeey: if you return a flag from `RemoveTaints`, here will be 
```
removed = helpers.RemoveTaints(&newTaints, UnavailableTaint, UnreachableTaint)
if removed {
 eventType = ...
 msg = ....
}
```

pkg/hub/taint/controller.go
```diff
@@ -0,0 +1,98 @@
...
+	case meta.IsStatusConditionFalse(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable):
+		helpers.RemoveTaints(&newTaints, UnreachableTaint)
+		helpers.AddTaints(&newTaints, UnavailableTaint)
+		c.eventRecorder.Eventf("ManagedClusterConditionAvailable conditionStatus is False", "change UnreachableTaint to UnavailableTaint")
```
skeeey: if you have a flag from `RemoveTaints` and `AddTaints` here will be 
```
if remvoed || udpaed {
 msg = ...
}
```

pkg/hub/taint/controller.go
elgnay: The cluster may have no taint at all. Please only create event only when a taint is removed from a cluster.

skeeey: And here you may only set the message, don't recode the event, because the update action may be failed, you need record the event unitl the update action success

pkg/hub/taint/controller.go
```diff
@@ -0,0 +1,98 @@
...
+	case meta.FindStatusCondition(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable) == nil ||
+		meta.IsStatusConditionPresentAndEqual(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable, metav1.ConditionUnknown):
+		helpers.RemoveTaints(&newTaints, UnavailableTaint)
+		helpers.AddTaints(&newTaints, UnreachableTaint)
+		c.eventRecorder.Eventf("ManagedClusterConditionAvailable conditionStatus is Unknown or without this condition",
+			"change UnavailableTaint to UnreachableTaint")
+	}
+
+	if helpers.IsTaintsEqual(managedCluster.Spec.Taints, newTaints) {
```
skeeey: if `RemoveTaints` and `AddTaints` have a flag, we may eliminate this judgement, am I right?

pkg/hub/taint/controller.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil
+	}
+	managedCluster.Spec.Taints = newTaints
+
+	if _, err = c.clusterClient.ClusterV1().ManagedClusters().Update(ctx, managedCluster, metav1.UpdateOptions{}); err != nil {
+		return err
+	}
```
skeeey, line 94: record event here

pkg/hub/taint/controller.go
```diff
@@ -0,0 +1,124 @@
...
+	managedCluster = managedCluster.DeepCopy()
+	newTaints := managedCluster.Spec.Taints
+	conditionStatus := c.getConditionStatus(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable)
+
+	switch {
+	case conditionStatus == "NoCondition" || conditionStatus == "Unknown":
+		newTaints, removed = helpers.RemoveTaints(newTaints, UnavailableTaint)
+		newTaints, updated = helpers.AddTaints(newTaints, UnreachableTaint)
+		if removed || updated {
+			reason = "ManagedClusterConditionAvailable conditionStatus is Unknown or without this condition"
```
elgnay: Normally, the content of `reason` is a code, like `NoAvailableCondition` or `UnknownAvailableCondition`, while the message contains detail information.

pkg/hub/taint/controller.go
```diff
@@ -0,0 +1,124 @@
...
+			msg = "Change UnavailableTaint to UnreachableTaint"
```
elgnay: How do you know the state is changed from `UnavailableTaint` to `UnreachableTaint`?

pkg/helpers/helpers.go
```diff
@@ -321,3 +321,61 @@ func ManagedClusterAssetFn(fs embed.FS, managedClusterName string) resourceapply
 		return assets.MustCreateAssetFromTemplate(name, template, config).Data, nil
 	}
 }
+
+func TaintEqual(taint1 clusterv1.Taint, taint2 clusterv1.Taint) bool {
```
elgnay: TaintEqual -> IsTaintEqual?

pkg/helpers/helpers.go
```diff
@@ -321,3 +321,61 @@ func ManagedClusterAssetFn(fs embed.FS, managedClusterName string) resourceapply
...
+		return false
+	}
+	if (taints1 == nil) != (taints2 == nil) {
+		return false
+	}
+	if len(taints2) == 0 && len(taints1) == 0 {
+		return true
+	}
+	for _, v1 := range taints1 {
+		flag := false
```
elgnay: flag -> found?

pkg/helpers/helpers.go
```diff
@@ -321,3 +321,61 @@ func ManagedClusterAssetFn(fs embed.FS, managedClusterName string) resourceapply
 		return assets.MustCreateAssetFromTemplate(name, template, config).Data, nil
 	}
 }
+
+func TaintEqual(taint1 clusterv1.Taint, taint2 clusterv1.Taint) bool {
+	// Ignore the comparison of time
+	return taint1.Key == taint2.Key && taint1.Value == taint2.Value && taint1.Effect == taint2.Effect
+}
+
+func IsTaintsEqual(taints1 []clusterv1.Taint, taints2 []clusterv1.Taint) bool {
```
elgnay: A better implementation of this function is to build an index for the first taint array. The key is that taint itself without the `timeAdded`. And then try to match each element in the second taint array with the index.

pkg/helpers/helpers.go
```diff
@@ -321,3 +321,61 @@ func ManagedClusterAssetFn(fs embed.FS, managedClusterName string) resourceapply
...
+		}
+	}
+
+	return true
+}
+
+func AddTaints(taints []clusterv1.Taint, taint clusterv1.Taint) (res []clusterv1.Taint, updated bool) {
+	for _, v := range taints {
+		if v.Key == taint.Key {
+			return taints, false
```
elgnay: You need to check the value/effect of taint before return

elgnay: And please always return a new array

pkg/helpers/helpers.go
```diff
@@ -321,3 +321,61 @@ func ManagedClusterAssetFn(fs embed.FS, managedClusterName string) resourceapply
...
+		}
+		if !flag {
+			return false
+		}
+	}
+
+	return true
+}
+
+func AddTaints(taints []clusterv1.Taint, taint clusterv1.Taint) (res []clusterv1.Taint, updated bool) {
```
elgnay: Please add doc to those helper functions

skeeey:
@JiahaoWei-RH you need rebase this pr

pkg/helpers/helpers.go
```diff
@@ -246,3 +246,57 @@ func FindTaintByKey(managedCluster *clusterv1.ManagedCluster, key string) *clust
 	}
 	return nil
 }
+
+func IsTaintEqual(taint1 clusterv1.Taint, taint2 clusterv1.Taint) bool {
```
qiujian16: func IsTaintEqual(taint1, taint2 clusterv1.Taint)

pkg/helpers/helpers.go
qiujian16: is this only used in testing? I feel you can use deep equal here.

pkg/helpers/helpers.go
```diff
@@ -246,3 +246,57 @@ func FindTaintByKey(managedCluster *clusterv1.ManagedCluster, key string) *clust
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// AddTaints add taints to the specified slice, if it did not already exist.
+// Return a new slice and a boolean indicating whether the slice has been updated.
+func AddTaints(taints []clusterv1.Taint, taint clusterv1.Taint) ([]clusterv1.Taint, bool) {
```
qiujian16: I think

```
func AddTaints(taints *[]clusterv1.Taint, taint clusterv1.Taint)
```

will make things easier

pkg/helpers/helpers.go
```diff
@@ -246,3 +246,57 @@ func FindTaintByKey(managedCluster *clusterv1.ManagedCluster, key string) *clust
...
+	for _, v := range taints {
+		if IsTaintEqual(v, taint) {
+			return res, false
+		}
+	}
+	res = append(res, taint)
+	return res, true
+}
+
+func RemoveTaints(taints []clusterv1.Taint, targets ...clusterv1.Taint) (res []clusterv1.Taint, updated bool) {
```
qiujian16: I think

```go
func RemoveTaints(taints *[]clusterv1.Taint, targets taint clusterv1.Taint)
```

will make things eaiser

pkg/hub/taint/controller.go
```diff
@@ -0,0 +1,121 @@
...
+		return err
+	}
+
+	var reason string
+	var msg string
+	var removed bool
+	var updated bool
+	managedCluster = managedCluster.DeepCopy()
+	newTaints := managedCluster.Spec.Taints
+	conditionStatus := c.getConditionStatus(managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable)
```
qiujian16: I would do like

```go
cond := meta.FindStatusCondition(&managedCluster.Status.Conditions, v1.ManagedClusterConditionAvailable)

switch {
case cond == nil || cond.Status == metav1.StatusUnknown:
  helpers.AddTaints(&managedCluster.Spec.Taints, UnreachableTaint)
  helpers.RemoveTaint(&managedCluster.Spec.Taints, UnavailableTaint)
case cond.Status==metav1.StatusFalse:
  helpers.AddTaints(&managedCluster.Spec.Taints, UnavailableTaint)
  helpers.RemoveTaint(&managedCluster.Spec.Taints, UnreachableTaint)
case cond.Status==metav1.StatusTrue:
  helpers.RemoveTaint(&managedCluster.Spec.Taints, UnavailableTaint)
  helpers.RemoveTaint(&managedCluster.Spec.Taints, UnreachableTaint)
}

// do a deepequal here or have a single updated(modified) flag.
```

pkg/helpers/helpers.go
JiahaoWei-RH: In the latest code, i didn' t remove this func because it is used in controller.

pkg/hub/taint/controller.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return err
+	}
+
+	managedCluster = managedCluster.DeepCopy()
+	newTaints := managedCluster.Spec.Taints
+	cond := managedCluster.Status.Conditions
+
+	switch {
+	case meta.IsStatusConditionPresentAndEqual(cond, v1.ManagedClusterConditionAvailable, metav1.ConditionUnknown) ||
```
qiujian16: I think you could

```go
cond := meta.FindStatusCondition(cond, v1.ManagedClusterConditionAvailable)
```

in case, maybe it could be

```
updated = helpers.RemoveTaints(&newTaints, UnavailableTaint)
updated = helpers.AddTaints(&newTaints, UnreachableTaint) && updated
```

then case based on cond.Status. Note that IsStatusConditionFalse always call FindStatusCondition. And since your removeTaints, addTatins return updated. I think you can do 

```
if updated
```
finally.

I think we are almost done. Thanks!

pkg/helpers/helpers.go
```diff
@@ -246,3 +246,57 @@ func FindTaintByKey(managedCluster *clusterv1.ManagedCluster, key string) *clust
...
+	}
+	*taints = append(*taints, taint)
+	return true
+}
+
+func RemoveTaints(taints *[]clusterv1.Taint, targets ...clusterv1.Taint) (updated bool) {
+	var newTaints []clusterv1.Taint
+	for _, v1 := range *taints {
+		found := false
+		for _, v2 := range targets {
```
qiujian16: you could make it another func in this for as
```
FindTaints(taints []clusterv1.Taint, taint clusterv1.Taint) *clusterv1.Taint
```
This could be used in both AddTaint and RemoveTaint

test/integration/taint_add_test.go
```diff
@@ -0,0 +1,148 @@
...
+			spokeCluster, err := util.GetManagedCluster(clusterClient, managedClusterName)
+			if err != nil {
+				return err
+			}
+			spokeCluster.Spec.HubAcceptsClient = true
+			_, err = clusterClient.ClusterV1().ManagedClusters().Update(context.TODO(), spokeCluster, metav1.UpdateOptions{})
+			return err
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.Succeed())
+
+		gomega.Eventually(func() bool {
```
qiujian16: nit: try to return err here, that will help you to debug

pkg/helpers/helpers.go
qiujian16: nit: remove this add use deepequal in your test.

qiujian16 (commented):
/approve

Thanks, just a nit comment to remove IsTaintsEqual helper.

elgnay:
One last comment: if a ManagedCluster starts terminating, do not update it (In another word, just ignore terminating clusters). Here is an example: https://github.com/open-cluster-management-io/registration/blob/main/pkg/hub/managedclusterset/controller.go#L133-L136

elgnay:
The related task: 
- https://github.com/stolostron/backlog/issues/18666
- https://github.com/stolostron/backlog/issues/19749

PR #201, opened 2022-02-24
set default value for clusterselector fields

ldpliu:


pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -42,22 +39,137 @@ func (a *ManagedClusterMutatingAdmissionHook) MutatingResource() (schema.GroupVe
...
+			Status: metav1.StatusFailure, Code: http.StatusBadRequest, Reason: metav1.StatusReasonBadRequest,
+			Message: err.Error(),
+		}
+		return nil, status
+	}
+
+	newClusterSet := clusterSet.DeepCopy()
+
+	if clusterSet.Spec.ClusterSelector.SelectorType == "" {
+		newClusterSet.Spec.ClusterSelector.SelectorType = clusterv1beta1.ExclusiveLabel
```
qiujian16, line 117: I do not think you need this, it is already set by default

pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -42,22 +39,137 @@ func (a *ManagedClusterMutatingAdmissionHook) MutatingResource() (schema.GroupVe
...
+	}
+
+	if clusterSet.Spec.ClusterSelector.ExclusiveLabel != nil {
```
qiujian16, line 120: this should be done in validatingwebhook, right?

pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -42,22 +39,137 @@ func (a *ManagedClusterMutatingAdmissionHook) MutatingResource() (schema.GroupVe
...
+			status.Allowed = false
+			status.Result = &metav1.Status{
+				Status: metav1.StatusFailure, Code: http.StatusInternalServerError, Reason: metav1.StatusReasonInternalError,
+				Message: fmt.Sprintf("The spec.clusterSelector.ExclusiveLabel.value must be %q.", clusterSet.Name),
+			}
+			return nil, status
+		}
+	}
+
+	newClusterSet.Spec.ClusterSelector.ExclusiveLabel = &clusterv1beta1.ManagedClusterLabel{
```
qiujian16, line 141: would you do a comparison here, and noop if this is already set

ldpliu:
close as it's not needed in this release

PR #202, opened 2022-02-24
add controller to set default clusterset label

ycyaoxdu:
add a controller to set default cluster set label for managed clusters which not set cluster set label.
the label will be `cluster.open-cluster-management.io/clusterset:default`

### DIFF
--- a/pkg/hub/managedclusterset/default_managedclusterset_label_controller.go
+++ b/pkg/hub/managedclusterset/default_managedclusterset_label_controller.go
@@ -1,0 +1,106 @@
+package managedclusterset
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	clientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	clusterinformerv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	clusterlisterv1 "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+const (
+	defaultManagedClusterSetValue = "default"
+)
+
+// defaultManagedClusterSetLabelController is used to add a "default" clusterset label to managedcluster
+// to add it to default managed cluster set if it it not belongs to any cluster set and have no cluster
+// set label.
+// This controller would be removed in next release.
+//
+// defaultManagedClusterSetLabelController reconciles ManagedClusterSet label.
+type defaultManagedClusterSetLabelController struct {
+	clusterClient clientset.Interface
+	clusterLister clusterlisterv1.ManagedClusterLister
+	eventRecorder events.Recorder
+}
+
+// NewDefaultManagedClusterSetLabelController creates a new set ManagedClusterSet label controller
+func NewDefaultManagedClusterSetLabelController(
+	clusterClient clientset.Interface,
+	clusterInformer clusterinformerv1.ManagedClusterInformer,
+	recorder events.Recorder) factory.Controller {
+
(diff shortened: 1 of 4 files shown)


ldpliu:
please change file name like: a_b_c

ldpliu:
I am not sure if it's better to put the logic to current managedcluster controller.(pkg/hub/managedcluster/controller.go)

pkg/hub/managedclusterset/setdefaultlabelcontroller.go
```diff
@@ -0,0 +1,94 @@
...
+	clusterlisterv1 "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+const (
+	defaultManagedClusterSetValue = "default"
+)
+
+// setDefaultManagedClusterSetLabelController reconciles ManagedClusterSet label.
+type setDefaultManagedClusterSetLabelController struct {
```
skeeey: how about defaultManagedClusterSetLabelController?

skeeey:
> I am not sure if it's better to put the logic to current managedcluster controller.(pkg/hub/managedcluster/controller.go)

I don't think it's a better way to put the logic to the managedcluster controller, using an independent controller make the logic more clear，and managedcluster controller is used to handle accept a managed cluster,

pkg/hub/managedclusterset/setdefaultlabelcontroller.go
ycyaoxdu: done

pkg/hub/managedclusterset/default_managedclusterset_label_controller.go
```diff
@@ -0,0 +1,94 @@
...
+	if err := c.syncClusterSetLabel(ctx, managedCluster); err != nil {
+		return fmt.Errorf("failed to set default ManagedClusterSet label to ManagedCluster %s: %w", managedCluster.Name, err)
+	}
+	return nil
+}
+
+func (c *defaultManagedClusterSetLabelController) syncClusterSetLabel(ctx context.Context, managedCluster *clusterv1.ManagedCluster) error {
+	cluster := managedCluster.DeepCopy()
+
+	if cluster.Labels == nil {
```
qiujian16: I would use resourcemerge.MergeLabels

qiujian16 (commented):
I think we should add a feature gate for this. Also as describe in the community meeting, we should have an annotation to indicate whether this should still be managed by registration.

pkg/hub/managedclusterset/default_managedclusterset_label_controller.go
```diff
@@ -0,0 +1,95 @@
...
+		return fmt.Errorf("failed to set default ManagedClusterSet label to ManagedCluster %s: %w", managedCluster.Name, err)
+	}
+	return nil
+}
+
+func (c *defaultManagedClusterSetLabelController) syncClusterSetLabel(ctx context.Context, managedCluster *clusterv1.ManagedCluster) error {
+	cluster := managedCluster.DeepCopy()
+	modified := false
+
+	clusterSetLabels := map[string]string{}
```
elgnay: Do not move all clusters into the default clusterset. Instead, please check if the cluster has already been a member of a clusterset first.

ycyaoxdu: done. Add if condition to check.

ycyaoxdu:
@elgnay thanks a lot for help :)

pkg/hub/managedclusterset/default_managedclusterset_label_controller.go
```diff
@@ -0,0 +1,101 @@
...
+	clusterinformerv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	clusterlisterv1 "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+const (
+	defaultManagedClusterSetValue = "default"
+)
+
+// defaultManagedClusterSetLabelController reconciles ManagedClusterSet label.
```
qiujian16, line 30: Add a comment here to indicate that the purpose of this controller is to add clusterset of the old version to the default clusterset if it has not cluster label. And this controller should be removed in the next release

ycyaoxdu: done. https://github.com/open-cluster-management-io/registration/pull/202/files#diff-c991f1a201d952eec15d4a0796413eae3706e1e41ea4144640fefb7e82b554faR25-R28

qiujian16 (commented):
/approve
/lgtm

PR #203, opened 2022-02-25
add label permission to join clusterset

ldpliu:

### DIFF
--- a/pkg/webhook/cluster/validating_webhook.go
+++ b/pkg/webhook/cluster/validating_webhook.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	"open-cluster-management.io/registration/pkg/helpers"
@@ -23,7 +24,9 @@ import (
 )
 
 const (
-	clusterSetLabel = "cluster.open-cluster-management.io/clusterset"
+	clusterSetLabel               = "cluster.open-cluster-management.io/clusterset"
+	internalClusterSetLabelPrefix = "cluster.open-cluster-management.io/"
+	infoClusterSetLabelPrefix     = "info.open-cluster-management.io/"
 )
 
 // ManagedClusterValidatingAdmissionHook will validate the creating/updating managedcluster request.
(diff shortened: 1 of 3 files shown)


pkg/webhook/cluster/validating_webhook.go
```diff
@@ -292,3 +291,74 @@ func (a *ManagedClusterValidatingAdmissionHook) allowUpdateClusterSet(userInfo a
...
+	if status := a.allowJoinClusterSet(userInfo, clusterSetName); status.Allowed {
+		return status
+	}
+	// check "managedclusters/label" permission with resource name "labelKey:labelValue"
+	if status := a.allowUpdateLabel(userInfo, clusterSetLabel+":"+clusterSetName); status.Allowed {
+		return status
+	}
+
+	// check "managedclusters/label" permission with resource name "labelKey:*"
+	if status := a.allowUpdateLabel(userInfo, clusterSetLabel+":*"); status.Allowed {
```
qiujian16: why we need this?

pkg/webhook/cluster/validating_webhook.go
```diff
@@ -292,3 +291,74 @@ func (a *ManagedClusterValidatingAdmissionHook) allowUpdateClusterSet(userInfo a
...
+}
+
+// allowUpdateClusterSet check where a request user has been authorized to add/remove the clustersetLabel to/from ManagedCluster
+func (a *ManagedClusterValidatingAdmissionHook) allowUpdateClusterSet(userInfo authenticationv1.UserInfo, clusterSetName string) *admissionv1beta1.AdmissionResponse {
+	// check "managedclustersets/join" permission
+	if status := a.allowJoinClusterSet(userInfo, clusterSetName); status.Allowed {
+		return status
+	}
+	// check "managedclusters/label" permission with resource name "labelKey:labelValue"
+	if status := a.allowUpdateLabel(userInfo, clusterSetLabel+":"+clusterSetName); status.Allowed {
```
qiujian16: to my memory, when check join we also check both if user can join old  and new clusterset.

pkg/webhook/cluster/validating_webhook.go
ldpliu: this is used to support permission like:
` - apiGroups: ["cluster.open-cluster-management.io"]
   resources: ["managedclusters/label"]
   resourceNames: ["cluster.open-cluster-management.io/clusterset:*"]
   verbs: ["create"]`
I am not sure if we should support it?

pkg/webhook/cluster/validating_webhook.go
ldpliu: Yes, we did the check in https://github.com/open-cluster-management-io/registration/pull/203/files#diff-042b72f0f7d4c234eb9f079b0fdfa71967b9f9c97dc0a121084177114ac3d1f6L226

This func is used to check one set's permission

pkg/webhook/cluster/validating_webhook.go
qiujian16: can you elaborate what '*' means here

pkg/webhook/cluster/validating_webhook.go
qiujian16: To my understanding, you should have a check func that look at the changed labels on the cluster, filter out the labels that is not used to do rbac check, then using the remaining labels to do rbac check

PR #206, opened 2022-03-02
update clusterset contoller to ignore non-legacy clusterset

elgnay:
Related task: https://github.com/stolostron/backlog/issues/20209

### DIFF
--- a/pkg/hub/managedclusterset/controller.go
+++ b/pkg/hub/managedclusterset/controller.go
@@ -141,8 +141,15 @@ func (c *managedClusterSetController) sync(ctx context.Context, syncCtx factory.
 	return nil
 }
 
-// syncClusterSet syncs a particular cluster set
+// syncClusterSet syncs a particular cluster set. Currently only support the legacy type of clusterset.
+// The logic should be updated once any new type of clusterset added.
 func (c *managedClusterSetController) syncClusterSet(ctx context.Context, originalClusterSet *clusterv1beta1.ManagedClusterSet) error {
+	// ignore the non-legacy clusterset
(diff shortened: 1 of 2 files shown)


pkg/hub/managedclusterset/controller.go
```diff
@@ -135,6 +135,12 @@ func (c *managedClusterSetController) sync(ctx context.Context, syncCtx factory.
 		return nil
 	}
 
+	// only hand the legacy clusterset
```
qiujian16: handle

pkg/hub/managedclusterset/controller.go
```diff
@@ -135,6 +135,12 @@ func (c *managedClusterSetController) sync(ctx context.Context, syncCtx factory.
...
+	selectorType := clusterSet.Spec.ClusterSelector.SelectorType
+	if len(selectorType) > 0 && selectorType != clusterv1beta1.LegacyClusterSetLabel {
```
qiujian16: This is still a bit strange to me, I think we should check the type in syncClusterSet, or we should at least add a comment indicating if more type is added in the future, this code should be removed.

pkg/hub/managedclusterset/controller.go
elgnay: fixed

pkg/hub/managedclusterset/controller.go
elgnay: Moved it into `syncClusterSet` as suggested.

qiujian16 (commented):
/approve
/lgtm

PR #209, opened 2022-03-07
fix override cluster labels

ldpliu:

### DIFF
--- a/pkg/webhook/cluster/mutating_webhook.go
+++ b/pkg/webhook/cluster/mutating_webhook.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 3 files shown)


pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -217,10 +217,10 @@ func newTaintTimeAddedJsonPatch(index int, timeAdded time.Time) jsonPatchOperati
 	}
 }
 
-func newLabelJsonPatch() jsonPatchOperation {
+func newLabelJsonPatch(labels map[string]string) jsonPatchOperation {
 	return jsonPatchOperation{
-		Operation: "add",
+		Operation: "replace",
```
zhujian7: If the original labels map is nil, will this work?

test/e2e/webhook_test.go
```diff
@@ -714,7 +715,11 @@ func newManagedCluster(name string, accepted bool, externalURL string) *clusterv
 	return &clusterv1.ManagedCluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
+			Labels: map[string]string{
```
zhujian7: Please cover cases labels nil as well as not nil.

pkg/webhook/cluster/mutating_webhook.go
ldpliu: it could not be nil,
As it must be has `clusterSetLabels[clusterSetLabel] = defaultClusterSetName`

test/e2e/webhook_test.go
ldpliu: it could not be nil,
As it must be has `clusterSetLabels[clusterSetLabel] = defaultClusterSetName`

pkg/webhook/cluster/mutating_webhook.go
zhujian7: I mean the **original** object, this func just returns a json patch, right？ where will the json patch be used? Could this patch apply to the ManagedCluster object(labels nil) successfully？

pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -133,7 +133,7 @@ func (a *ManagedClusterMutatingAdmissionHook) addDefaultClusterSetLabel(managedC
 		return nil, status
 	}
 
-	labelPatch := newLabelJsonPatch()
+	labelPatch := newLabelJsonPatch(cluster.Labels)
```
qiujian16: if it is add operation, why it needs to know the original label?

ldpliu: The `add`  operation for `labels` will override the existing `labels`.

qiujian16 (commented):
/approve

PR #210, opened 2022-03-07
add disable annotation detection for default managed cluster set

ycyaoxdu:
add disable annotation detection for default managed cluster set, ensuring users have property to modify default managed cluster set spec

### DIFF
--- a/pkg/hub/managedclusterset/default_managedclusterset_controller.go
+++ b/pkg/hub/managedclusterset/default_managedclusterset_controller.go
@@ -19,7 +19,10 @@ import (
 	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
 )
 
-const defaultManagedClusterSetName = "default"
+const (
+	autoUpdateAnnotation         = "cluster.open-cluster-management.io/autoupdate"
+	defaultManagedClusterSetName = "default"
+)
 
 var defaultManagedClusterSetSpec = clusterv1beta1.ManagedClusterSetSpec{
 	ClusterSelector: clusterv1beta1.ManagedClusterSelector{
@@ -110,6 +113,12 @@ func (c *defaultManagedClusterSetController) sync(ctx context.Context, syncCtx f
 func (c *defaultManagedClusterSetController) syncDefaultClusterSet(ctx context.Context, originalDefaultClusterSet *clusterv1beta1.ManagedClusterSet) error {
 	defaultClusterSet := originalDefaultClusterSet.DeepCopy()
 
+	// if the annotation has set to disable, default clusterset controller will not work.
+	if hasAnnotation(defaultClusterSet, autoUpdateAnnotation, "false") {
+		klog.V(4).Info("DefaultManagedClusterSetDisabled", "The DefaultManagedClusterSet is disabled by user")
+		return nil
+	}
+
 	// if defaultClusterSet.Spec is changed, rollback the change by update it to the original value.
 	if !equality.Semantic.DeepEqual(defaultClusterSet.Spec, defaultManagedClusterSetSpec) {
 		defaultClusterSet.Spec = defaultManagedClusterSetSpec
@@ -123,4 +132,14 @@ func (c *defaultManagedClusterSetController) syncDefaultClusterSet(ctx context.C
 	}
 
 	return nil
+}
+
+func hasAnnotation(set *clusterv1beta1.ManagedClusterSet, key, value string) bool {
+	if set.Annotations == nil {
+		return false
+	}
+	if v, ok := set.Annotations[key]; ok && v == value {
+		return true
+	}
+	return false
 }
--- a/test/integration/default_managedclusterset_controller_test.go
+++ b/test/integration/default_managedclusterset_controller_test.go
@@ -60,4 +60,6 @@ var _ = ginkgo.Describe("DefaultManagedClusterSet", func() {
 		}, eventuallyTimeout, eventuallyInterval).ShouldNot(gomega.HaveOccurred())
 	})
 
+	// TODO:
+	//		add test case for edit default managedclusterset spec operation
 })
--- a/pkg/hub/managedclusterset/default_managedclusterset_controller_test.go
+++ b/pkg/hub/managedclusterset/default_managedclusterset_controller_test.go
@@ -67,6 +67,13 @@ func TestSyncDefaultClusterSet(t *testing.T) {
 				}
 			},
 		},
+		{
+			name:               "sync default cluster set with disabled annotation",
+			existingClusterSet: newDefaultManagedClusterSetWithAnnotation(defaultManagedClusterSetName, autoUpdateAnnotation, "false", defaultManagedClusterSetSpec, false),
+			validateActions: func(t *testing.T, actions []clienttesting.Action) {
+				testinghelpers.AssertNoActions(t, actions)
+			},
+		},
 	}
 
 	for _, c := range cases {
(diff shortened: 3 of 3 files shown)


ycyaoxdu:
@qiujian16  @ldpliu  please help to review.

pkg/hub/managedclusterset/default_managedclusterset_controller.go
```diff
@@ -110,6 +114,13 @@ func (c *defaultManagedClusterSetController) sync(ctx context.Context, syncCtx f
 func (c *defaultManagedClusterSetController) syncDefaultClusterSet(ctx context.Context, originalDefaultClusterSet *clusterv1beta1.ManagedClusterSet) error {
 	defaultClusterSet := originalDefaultClusterSet.DeepCopy()
 
+	// if the annotation has set to disable, default clusterset controller will not work.
+	if hasAnnotation(defaultClusterSet, defaultClusterSetAnnotationKey, defaultClusterSetAnnotationDisabledValue) {
+		c.eventRecorder.Eventf("DefaultManagedClusterSetDisabled", "The DefaultManagedClusterSet is disabled by user")
```
qiujian16: I think it will generate a lot of event if annotation is set. We may only set a log with v(4)

pkg/hub/managedclusterset/default_managedclusterset_controller.go
```diff
@@ -19,7 +19,11 @@ import (
 	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
 )
 
-const defaultManagedClusterSetName = "default"
+const (
+	defaultClusterSetAnnotationKey           = "open-cluster-management.io/defaultclusterset"
```
qiujian16: I think we should check what is the current annotation set by bootstrap rbac to make it unmanaged. We should use the similar way. Note we probably have other bootstrapped clusterset in the future

ycyaoxdu: > I think we should check what is the current annotation set by bootstrap rbac to make it unmanaged. We should use the similar way. Note we probably have other bootstrapped clusterset in the future

found a annotation: [`DeploymentIgnorePodAnnotation = "deploy.openshift.io/deployer-pod.ignore"`](https://github.com/openshift/library-go/blob/master/pkg/apps/appsutil/const.go#L51), so i define the annotation like this: 
[`"open-cluster-management.io/managedclusterset.default.unmanage"`](https://github.com/open-cluster-management-io/registration/pull/210/files#diff-83468281441062b8e876bcef48f61044f897afce081d9790253027dacd34fc78R23) 
if this annotation exists on default clusterset, we will unmanage clusterset, just focus on the annotation key.

ycyaoxdu: > I think we should check what is the current annotation set by bootstrap rbac to make it unmanaged. We should use the similar way. Note we probably have other bootstrapped clusterset in the future

Not sure if I got it correctly. @qiujian16 please help to review

ldpliu: I think we should use the annotation like: `open-cluster-management.io/autoupdate:false` which is same as https://kubernetes.io/docs/reference/access-authn-authz/rbac/#auto-reconciliation

ycyaoxdu: > I think we should use the annotation like: `open-cluster-management.io/autoupdate:false` which is same as https://kubernetes.io/docs/reference/access-authn-authz/rbac/#auto-reconciliation

will we set `open-cluster-management.io/autoupdate:true` if we want it to be reconciled?

ldpliu: Same as the rbac logic here: https://github.com/kubernetes/kubectl/blob/0f88fc6b598b7e883a391a477215afb080ec7733/pkg/cmd/auth/reconcile.go
We may just need to check if this annotation exist `open-cluster-management.io/autoupdate:false`

We do not set it, and if user want to stop reconcile, he need to set  `open-cluster-management.io/autoupdate:false` to the default set.

ycyaoxdu: @ldpliu how about this? https://github.com/open-cluster-management-io/registration/pull/210/files#diff-83468281441062b8e876bcef48f61044f897afce081d9790253027dacd34fc78R23-R24

pkg/hub/managedclusterset/default_managedclusterset_controller.go
```diff
@@ -19,7 +19,11 @@ import (
 	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
 )
 
-const defaultManagedClusterSetName = "default"
+const (
+	defaultClusterSetAnnotation    = "open-cluster-management.io/autoupdate"
+	defaultClusterSetDisabledValue = "false"
```
ldpliu: Just as @qiujian16 Sailed, we probably have other bootstrapped clusterset in the future. And this annotation may be used for other sets.
So I suggest change the var name as: `autoUpdateAnnotation` and there is no need to define the value. just use `false` in code logic.

pkg/hub/managedclusterset/default_managedclusterset_controller.go
ldpliu: yes, the code logic looks ok for me, and give another comment about variable name.

pkg/hub/managedclusterset/default_managedclusterset_controller.go
ycyaoxdu: done https://github.com/open-cluster-management-io/registration/pull/210/files#diff-83468281441062b8e876bcef48f61044f897afce081d9790253027dacd34fc78R23

pkg/hub/managedclusterset/default_managedclusterset_controller.go
```diff
@@ -19,7 +19,10 @@ import (
 	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
 )
 
-const defaultManagedClusterSetName = "default"
+const (
+	autoUpdateAnnotation         = "open-cluster-management.io/autoupdate"
```
qiujian16: let's use cluster.open-cluser-management.io/autoupdate

pkg/hub/managedclusterset/default_managedclusterset_controller.go
```diff
@@ -124,3 +133,10 @@ func (c *defaultManagedClusterSetController) syncDefaultClusterSet(ctx context.C
 
 	return nil
 }
+
+func hasAnnotation(set *clusterv1beta1.ManagedClusterSet, key, value string) bool {
+	if v, ok := set.Annotations[key]; ok && v == value {
```
qiujian16, line 141: should also check set.Annotation==nil?

ycyaoxdu: done

test/integration/default_managedclusterset_controller_test.go
```diff
@@ -60,4 +60,6 @@ var _ = ginkgo.Describe("DefaultManagedClusterSet", func() {
 		}, eventuallyTimeout, eventuallyInterval).ShouldNot(gomega.HaveOccurred())
 	})
 
+	// TODO:
```
qiujian16, line 63: do you want to add test here?

ycyaoxdu: Actually i wrote some case to test whether the spec will be rollback is changed, but seems like we can't do this now. 
```
ManagedClusterSet.cluster.open-cluster-management.io "default" is invalid: spec.clusterSelector.selectorType: Unsupported value: "anything": supported values: "LegacyClusterSetLabel"
  occurred
```
So i just add the TODO comment here, warning that there is no integration test case for changing spec.

PR #214, opened 2022-03-17
V1beta1 CSR compatibility

yue9944882:


qiujian16 (commented):
It seems the only diff is

syncFunc()
resetFunc()
and the way to return informer(). 

How about we have a interface that

type csrControl interface {
    sync(csrName string, keyData []byte) error
    reset() error
    informer() cachedSharedIndedInformer
}

yue9944882:
@qiujian16 factored out a `csrControl` interface for transparent v1/v1beta1 csr api manipulation

pkg/clientcert/controller_test.go
```diff
@@ -183,10 +183,13 @@ func TestSync(t *testing.T) {
 			controller := &clientCertificateController{
 				ClientCertOption: clientCertOption,
 				CSROption:        csrOption,
-				hubCSRLister:     hubInformerFactory.Certificates().V1().CertificateSigningRequests().Lister(),
-				hubCSRClient:     hubKubeClient.CertificatesV1().CertificateSigningRequests(),
-				spokeCoreClient:  agentKubeClient.CoreV1(),
-				controllerName:   "test-agent",
+				csrControl: &v1CSRControl{
```
qiujian16: I think you would need some additional test for v1beta1 csr.

pkg/clientcert/certificate.go
```diff
@@ -157,16 +166,81 @@ func BuildKubeconfig(clientConfig *restclient.Config, certPath, keyPath string)
 	return kubeconfig
 }
 
-// isCSRApproved returns true if the given csr has been approved
-func isCSRApproved(csr *certificatesv1.CertificateSigningRequest) bool {
+type csrControl interface {
+	get(name string) (metav1.Object, error)
+	create(ctx context.Context, recorder events.Recorder, objMeta metav1.ObjectMeta, csrData []byte, signerName string) (string, error)
+	isApproved(csr metav1.Object) bool
```
qiujian16: should this be?

```
isApprovedandIssued(csrName string) ([]byte, bool, error)

```

and we do not need to expose `get`

qiujian16:
@elgnay @skeeey  this is a pretty critical change. Please also take a look at this PR.

pkg/clientcert/cert_controller.go
```diff
@@ -109,8 +108,49 @@ type clientCertificateController struct {
...
+			return nil, errors.Wrapf(err, "failed CSR api discovery")
+		}
+		if !v1CSRSupported && v1beta1CSRSupported {
+			csrCtrl = &v1beta1CSRControl{
+				hubCSRInformer: hubCSRInformer.V1beta1().CertificateSigningRequests(),
+				hubCSRLister:   hubCSRInformer.V1beta1().CertificateSigningRequests().Lister(),
+				hubCSRClient:   hubKubeClient.CertificatesV1beta1().CertificateSigningRequests(),
+			}
+		}
+		klog.Info("Using v1beta1 CSR api to manage spoke client certificate")
```
skeeey: This log should be in the `if !v1CSRSupported && v1beta1CSRSupported {`  block

pkg/clientcert/cert_controller.go
```diff
@@ -166,7 +205,47 @@ func (c *clientCertificateController) sync(ctx context.Context, syncCtx factory.
...
-		newSecretConfig, err := c.syncCSR()
+		// build a secret data map if the csr is approved
+		newSecretConfig, err := func() (map[string][]byte, error) {
+			// skip if there is no ongoing csr
+			if len(c.csrName) == 0 {
+				return nil, fmt.Errorf("no ongoing csr")
+			}
+
+			csr, err := c.csrControl.get(c.csrName)
+			if err != nil {
```
skeeey, line 216: if the csr is not found, why we remove the old fallback logic here? has some special reasons or we should keep the logic

pkg/clientcert/certificate.go
```diff
@@ -157,16 +166,81 @@ func BuildKubeconfig(clientConfig *restclient.Config, certPath, keyPath string)
...
+	isIssued(csr metav1.Object) ([]byte, bool)
```
skeeey, line 173: it seems that we get the certificate  from csr status in both v1 and v1beta1, what the different is between them? we may do not need this

pkg/clientcert/cert_controller.go
```diff
@@ -166,7 +205,47 @@ func (c *clientCertificateController) sync(ctx context.Context, syncCtx factory.
...
+				return nil, err
+			}
+
+			// skip if csr is not approved yet
+			if !c.csrControl.isApproved(csr) {
+				return nil, nil
+			}
+
+			// skip if csr has no certificate in its status yet
+			certData, issued := c.csrControl.isIssued(csr)
```
skeeey: nit: change the current comment to: skip if csr is not issued

pkg/clientcert/controller_test.go
yue9944882: added more tests, PTAL

pkg/clientcert/certificate.go
yue9944882: aesthetically i think separating isApproved and isIssued will look cleaner, but i eliminated getter from interface wdyt?

pkg/clientcert/cert_controller.go
yue9944882: nice catch

pkg/clientcert/cert_controller.go
yue9944882: done

yue9944882 (commented):
@qiujian16 i decoupled dependency for v1 csr api from the unit-tests to prove that the new csrControl interface is valid. and i wasnt able to add more integration/e2e tests b/c it will require running a 1.16 cluster in the github-action environment, but i enabled the feature-gates during e2e tests to prove the backward-compatibility

qiujian16:
/approve

I think it looks good. Ideally we might need another e2e test action for kube 1.16 to test this feature (if we need to support kube 1.16), but since it does not impact the existing code. we can do it as a follow up.

PR #215, opened 2022-03-22
Disable apf in webhook

qiujian16:

### DIFF
--- a/deploy/webhook/deployment.yaml
+++ b/deploy/webhook/deployment.yaml
@@ -24,6 +24,8 @@ spec:
           - "webhook"
           - "--cert-dir=/tmp"
           - "--secure-port=6443"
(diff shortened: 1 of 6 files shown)


deploy/webhook/deployment.yaml
```diff
@@ -24,6 +24,7 @@ spec:
           - "webhook"
           - "--cert-dir=/tmp"
           - "--secure-port=6443"
+          - "--feature-gates=DefaultClusterSet=true,APIPriorityAndFairness=false"
```
yue9944882, line 28: nit: clarify that admission webhook is not hosting any actual kubernetes resources, so it isnt subject to flowcontrol by APF.

yue9944882, line 28: interesting, it's an alpha feature but we're enabling it by default

UPDATE: nvmd, it's the test deploy script

PR #222, opened 2022-04-06
set clusterset to default when value is null string

ldpliu:
https://github.com/stolostron/backlog/issues/21335

### DIFF
--- a/pkg/webhook/cluster/mutating_webhook.go
+++ b/pkg/webhook/cluster/mutating_webhook.go
@@ -128,18 +128,34 @@ func (a *ManagedClusterMutatingAdmissionHook) addDefaultClusterSetLabel(managedC
 		return jsonPatches, status
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -128,7 +128,19 @@ func (a *ManagedClusterMutatingAdmissionHook) addDefaultClusterSetLabel(managedC
 		return jsonPatches, status
 	}
 
-	if _, ok := managedCluster.Labels[clusterSetLabel]; ok {
+	if v, ok := managedCluster.Labels[clusterSetLabel]; ok {
```
qiujian16: I would do

```
v, ok := managedCluster.Labels[clusterSetLabel]
if !ok {
  // add label
}

if len(v) == 0 {
  // update label
}

return nil
```

ldpliu: done

PR #224, opened 2022-04-06
allow ManagedCluster creation with taint.timeAdded specified

elgnay:


pkg/webhook/cluster/mutating_webhook.go
```diff
@@ -71,13 +71,15 @@ func (a *ManagedClusterMutatingAdmissionHook) Admit(req *admissionv1beta1.Admiss
...
 	var jsonPatches []jsonPatchOperation
+	var warnings []string
 
 	// set timeAdded of taint if it is nil and reset it if it is modified
 	taintJsonPatches, status := a.processTaints(managedCluster, req.OldObject.Raw)
 	if !status.Allowed {
 		return status
 	}
 	jsonPatches = append(jsonPatches, taintJsonPatches...)
+	warnings = append(warnings, status.Warnings...)
```
qiujian16: why we need to do append here?

elgnay: Redundant code removed.

PR #234, opened 2022-04-11
fix clusterset sync issue

elgnay:

### DIFF
--- a/pkg/hub/managedclusterset/controller.go
+++ b/pkg/hub/managedclusterset/controller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"reflect"
-	"sync"
 
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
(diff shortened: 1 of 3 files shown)


pkg/hub/managedclusterset/controller.go
```diff
@@ -177,46 +148,31 @@ func (c *managedClusterSetController) syncClusterSet(ctx context.Context, origin
...
-// given cluster set
-func (c *managedClusterSetController) updateClusterSetsMap(clusterSetName string, clusters []*clusterv1.ManagedCluster) {
-	c.mapLock.Lock()
-	defer c.mapLock.Unlock()
-
-	newMembers := sets.NewString()
-	for _, cluster := range clusters {
-		newMembers.Insert(cluster.Name)
+func getClusterSet(obj interface{}) string {
+	accessor, _ := meta.Accessor(obj)
```
qiujian16: you would need to handle tombstone case

qiujian16 (commented):
I think we should add an integration test for this

pkg/hub/managedclusterset/controller.go
elgnay: Done.

elgnay:
> I think we should add an integration test for this

We have already had the integration test case to cover this issue, but the out-of-sync happens occasionally. I update the test case to repeat the same testing 10 times.

PR #245, opened 2022-05-17
[WIP] Refactor cluster deletion process

qiujian16:
1. monitor addon/work when deletion
2. delete ns when a special label is not set
3. allow adding other resource to monitor before starting deletion process

### DIFF
--- a/pkg/hub/managedcluster/deletion_controller.go
+++ b/pkg/hub/managedcluster/deletion_controller.go
@@ -1,0 +1,303 @@
+package managedcluster
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/metadata"
+	"k8s.io/klog/v2"
+	addonv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	clientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	informerv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	listerv1 "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	v1 "open-cluster-management.io/api/cluster/v1"
+	workapiv1 "open-cluster-management.io/api/work/v1"
+	"open-cluster-management.io/registration/pkg/helpers"
+)
+
+const (
+	estimate                   = 3 * time.Second
+	ConditionTypeDeleteSuccess = "ContentDeleteSuccess"
+	ResourceRemainReason       = "ResourceRemaining"
+	FinalizerRemainReason      = "FinalizerRemaining"
+
+	// DeletionByOtherLabelKey is the key on resource, the resource will not be delete by registration
+	// with this key
+	DeletionByOtherLabelKey = "cluster.open-cluster-management.io/delete-by-other"
+)
+
+// managedClusterController reconciles instances of ManagedCluster on the hub.
+type managedClusterDeletionController struct {
+	kubeClient     kubernetes.Interface
+	clusterClient  clientset.Interface
+	metadataClient metadata.Interface
+	clusterLister  listerv1.ManagedClusterLister
+	eventRecorder  events.Recorder
+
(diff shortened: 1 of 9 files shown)


pkg/hub/managedcluster/deletion_controller.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	// delete all managedcluster addons.
+	remainingAddon, err := c.cleanupGVR(ctx, managedClusterName, addonv1alpha1.GroupVersion.WithResource("managedclusteraddons"), metav1.ListOptions{})
+	if err != nil || remainingAddon.numRemainingResource > 0 {
+		return remainingAddon, err
+	}
+
+	// delete all manifestworks
+	remainingWorks, err := c.cleanupGVR(ctx, managedClusterName, workapiv1.GroupVersion.WithResource("manifestworks"), metav1.ListOptions{
```
zhiweiyin318, line 198: I think we cannot delete all manifestWorks here if there is  addon pre-delete manifestWork.

pkg/hub/managedcluster/deletion_controller.go
```diff
@@ -0,0 +1,303 @@
...
+
+	// monitor all manifestworks again, this is to ensure all works have been deleted.
+	remainingWorks, err = c.monitorGVR(ctx, managedClusterName, workapiv1.GroupVersion.WithResource("manifestworks"), metav1.ListOptions{})
+	if err != nil || remainingWorks.numRemainingResource > 0 {
+		return remainingAddon, err
+	}
+
+	// for namespace deletion, we only delete ns with certain name and no deleteByOther label.
+	remainingNS, err := c.cleanupGVR(ctx, metav1.NamespaceAll, corev1.SchemeGroupVersion.WithResource("namespaces"), metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("!%s", DeletionByOtherLabelKey),
```
zhiweiyin318, line 215: should we delete the cluster ns if the cluster ns does not have the DeletionByOtherLabel, but there is a manifestWork with DeletionByOtherLabel.

pkg/hub/managedcluster/deletion_controller.go
```diff
@@ -0,0 +1,303 @@
...
+		FieldSelector: fmt.Sprintf("metadata.name=%s", managedClusterName),
+	})
```
zhiweiyin318, line 217: should we delete the rolebidings of registration/work agent if the cluster namespace has DeletionByOtherLabel?

pkg/hub/managedcluster/deletion_controller.go
qiujian16: comment is not correct, it will skip the work with a certain label

pkg/hub/managedcluster/deletion_controller.go
qiujian16: we will not until all work is removed, there is a monitor before this line

pkg/hub/managedcluster/deletion_controller.go
qiujian16: I think so, we do not want that permission exists when cluster is deleted.

pkg/hub/managedcluster/deletion_controller.go
zhiweiyin318: I think we also need to add the label for the manifestWorks created by addon, right?

pkg/hub/managedcluster/deletion_controller.go
zhiweiyin318: gotcha

pkg/hub/managedcluster/deletion_controller.go
```diff
@@ -0,0 +1,303 @@
...
+		return remainingAddon, err
+	}
+
+	// delete all manifestworks
+	remainingWorks, err := c.cleanupGVR(ctx, managedClusterName, workapiv1.GroupVersion.WithResource("manifestworks"), metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("!%s", DeletionByOtherLabelKey),
+	})
+
+	if err != nil || remainingWorks.numRemainingResource > 0 {
+		return remainingAddon, err
```
skeeey: `return remainingWorks, err`

pkg/hub/manager.go
```diff
@@ -74,6 +81,15 @@ func RunControllerManager(ctx context.Context, controllerContext *controllercmd.
 		controllerContext.EventRecorder,
 	)
 
+	managedClusterDeletionController := managedcluster.NewManagedClusterDeletionController(
+		kubeClient,
+		metaClient,
+		clusterClient,
+		clusterInformers.Cluster().V1().ManagedClusters(),
+		[]schema.GroupVersionResource{},
```
skeeey: we may consider to export this as a controller option, right?

pkg/hub/managedcluster/deletion_controller.go
skeeey: yeah, we need this label on the works that will be removed by other controller

pkg/hub/manager.go
qiujian16: yes, this should be under a feature gate also

pkg/hub/managedcluster/deletion_controller.go
qiujian16: addon will be call cleaned before start deleting manifestwork, so it should be fine.

PR #246, opened 2022-05-18
Add client cert condition when cert is rotated

qiujian16:

### DIFF
--- a/pkg/clientcert/cert_controller.go
+++ b/pkg/clientcert/cert_controller.go
@@ -42,6 +42,17 @@ const (
 
 	ClusterNameLabel = "open-cluster-management.io/cluster-name"
 	AddonNameLabel   = "open-cluster-management.io/addon-name"
+
+	// ClusterCertificateRotatedCondition is a condition type that client certificate is rotated
+	ClusterCertificateRotatedCondition = "ClusterCertificateRotated"
+
+	// ClientCertificateUpdateFailedReason is a reason of condition ClusterCertificateRotatedCondition that
+	// the client certificate rotation fails.
+	ClientCertificateUpdateFailedReason = "ClientCertificateUpdateFailed"
+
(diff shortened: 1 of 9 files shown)


pkg/spoke/managedcluster/registration.go
```diff
@@ -83,11 +87,29 @@ func NewClientCertForHubController(
 		spokeSecretInformer,
 		spokeKubeClient,
 		hubKubeClient,
+		statusUpdater,
 		recorder,
 		controllerName,
 	)
 }
 
+func GenerateBootStrapStatusUpdater() clientcert.StatusUpdateFunc {
```
elgnay: nit: GenerateBootStrapStatusUpdater -> GenerateBootstrapStatusUpdater?

pkg/clientcert/cert_controller.go
```diff
@@ -258,8 +285,43 @@ func (c *clientCertificateController) sync(ctx context.Context, syncCtx factory.
...
 			return err
 		}
+
+		timeBefore, timeAfter, err := getCertValidityPeriod(secret)
+
+		cond := metav1.Condition{
+			Type:    "ClusterCertificateRotated",
+			Status:  metav1.ConditionTrue,
+			Reason:  "ClientCertificateUpdated",
+			Message: fmt.Sprintf("client certificate rotated starting from %v to %v", *timeAfter, *timeBefore),
```
elgnay: Need to swap timeAfter and timeBefore in the argument list.

qiujian16: i thought timeAfter means cert is valid after timeAfter?

elgnay: I checked the code one more time and found `getCertValidityPeriod(secret)` returns `notBefore`, `notAfter` and `error`. So the validity period of a cert should start from `notBefore` to `notAfter`. What do you think?

PR #249, opened 2022-05-30
update qps and burst for webhook

ldpliu:

### DIFF
--- a/pkg/cmd/webhook/webhook.go
+++ b/pkg/cmd/webhook/webhook.go
@@ -5,6 +5,7 @@ import (
 
 	admissionserver "github.com/openshift/generic-admission-server/pkg/cmd/server"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	clusterwebhook "open-cluster-management.io/registration/pkg/webhook/cluster"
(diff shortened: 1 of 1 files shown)


qiujian16 (commented):
maybe we should put this as flag and set some default value

ldpliu:
> maybe we should put this as flag and set some default value

it's not easy to add the qps and burst flag for webhook. 
As current webhook use the `o.RecommendedOptions`
 https://github.com/open-cluster-management-io/registration/blob/fd33e20727c4d295f49b18c249d279369426c230/pkg/cmd/webhook/webhook.go#L46

qiujian16:
I think you can create another AddFlags options for this, right?

ldpliu:
> I think you can create another AddFlags options for this, right?

Done, please help to  review again.

pkg/cmd/webhook/webhook.go
```diff
@@ -44,6 +46,42 @@ func NewAdmissionHook() *cobra.Command {
...
+	if err != nil {
+		return err
+	}
+	return server.GenericAPIServer.PrepareRun().Run(stopCh)
+}
+
+// Config contains the server (the webhook) cert and key.
+type Options struct {
+	QPS   float32
+	Burst int
```
qiujian16: I would put adminssonServerOption in this Option. So all addFlags can be done in options.AddFlat.

qiujian16 (commented):
/approve
/lgtm

this is also needed for work

PR #250, opened 2022-06-02
remove add default set label controller

ldpliu:


qiujian16 (commented):
we should also remove the rbac rules, right?

ldpliu:
> we should also remove the rbac rules, right?

Removed https://github.com/open-cluster-management-io/registration/pull/250/files#diff-68752c39408458ef5dc35f2232378e0f2848f1ff035af41ab6a5fa57eb5831acL64
And will submit another pr to remove this in registration-oprator after this pr merged.

PR #256, opened 2022-06-23
Support install addons in customized namespaces

xuezhaojun:
Signed-off-by: xuezhaojun <zxue@redhat.com>

Epic: 
https://github.com/stolostron/backlog/issues/23093

User-story:
https://github.com/stolostron/backlog/issues/23929


pkg/spoke/addon/namespace_controller.go
```diff
@@ -0,0 +1,119 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	addoninformerv1alpha1 "open-cluster-management.io/api/client/addon/informers/externalversions/addon/v1alpha1"
+	addonlisterv1alpha1 "open-cluster-management.io/api/client/addon/listers/addon/v1alpha1"
+)
+
+const (
+	addonInstallNamespace = "addon.open-cluster-management.io/namespace"
```
qiujian16: addonInstallNamespaceAnnotationKey

pkg/spoke/addon/namespace_controller.go
```diff
@@ -0,0 +1,119 @@
...
+				_, err = c.kubeClient.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{})
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+
+	// When addon is deleted, check if there is anyother addon is using the same installNamespace, if not, delete the namespace
+	if !addOn.DeletionTimestamp.IsZero() {
```
qiujian16, line 78: I am worried about deleting ns here.

pkg/spoke/addon/namespace_controller.go
```diff
@@ -0,0 +1,119 @@
...
+) factory.Controller {
+	c := &addonNamespaceController{
+		kubeClient:  kubeClient,
+		addOnLister: addOnInformer.Lister(),
+		recorder:    recorder,
+	}
+	return factory.New().WithInformersQueueKeyFunc(func(o runtime.Object) string {
+		accessor, _ := meta.Accessor(o)
+		return accessor.GetName()
+	}, addOnInformer.Informer()).WithSync(c.sync).ToController("AddonNamespaceController", recorder)
```
qiujian16, line 46: should also watch ns also right? Also you would need registration to have permission to create ns. And update permission in registration-operator at first.

pkg/spoke/addon/namespace_controller.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	if err != nil {
+		return err
+	}
+
+	// Get installNamespace of managedClusterAddon
+	installNamespace := addOn.Spec.InstallNamespace
+
+	// Check installNamespace exist or not
+	ns, err := c.kubeClient.CoreV1().Namespaces().Get(ctx, installNamespace, metav1.GetOptions{})
```
qiujian16: use resourceApply.ApplyNamespace

qiujian16 (commented):
I do not think we should always delete the ns when addon is missing. Maybe we should have another annotation on ns so ns will not be deleted.

pkg/spoke/addon/namespace_controller.go
xuezhaojun: Fixed.

pkg/spoke/addon/namespace_controller.go
xuezhaojun: the permission is updated in this PR https://github.com/open-cluster-management-io/registration-operator/pull/253 . But not sure about watch ns.

pkg/spoke/addon/namespace_controller.go
xuezhaojun: Changed to use resourceapply pkg.

pkg/spoke/addon/namespace_controller.go
```diff
@@ -0,0 +1,95 @@
...
+				addonInstallNamespaceAnnotationKey: "true",
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	// When addon is deleted, check if there is anyother addon is using the same installNamespace, if not, delete the namespace
+	if !addOn.DeletionTimestamp.IsZero() {
```
skeeey, line 78: we may put this at the beginning

pkg/spoke/addon/namespace_controller.go
```diff
@@ -0,0 +1,95 @@
...
+		addOnList, err := c.addOnLister.ManagedClusterAddOns(c.managedClusterName).List(labels.Everything())
+		if err != nil {
+			return err
+		}
+		for _, a := range addOnList {
```
skeeey, line 83: exclude current addon?

qiujian16:
/hold 
since we had some further discussion. This may not be needed.

xuezhaojun:
/close
Change to the approach that addons are responsible for creating ns they need.

PR #265, opened 2022-07-20
recreate the hub kubeconfig when addon deploy option changes

zhujian7:

### DIFF
--- a/pkg/spoke/addon/configuration.go
+++ b/pkg/spoke/addon/configuration.go
@@ -21,9 +21,8 @@ const (
 // registrationConfig contains necessary information for addon registration
 // TODO: Refactor the code here once the registration configuration is available in spec of ManagedClusterAddOn
 type registrationConfig struct {
-	addOnName             string
-	installationNamespace string
-	registration          addonv1alpha1.RegistrationConfig
+	addOnName    string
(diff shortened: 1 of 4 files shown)


zhujian7:
cc @elgnay @qiujian16

elgnay:
nit: It might be better to include `installationNamespace` and `addOnAgentRunningOutsideManagedCluster` when calculating the hash of the registrationConfig. Then we can keep the logic of the registration controller neat.

qiujian16:
yes, I agree with @elgnay putting this into hash is better.

pkg/spoke/addon/configuration.go
```diff
@@ -34,6 +34,11 @@ type registrationConfig struct {
 	addOnAgentRunningOutsideManagedCluster bool
 }
 
+type addonInstallOption struct {
+	InstallationNamespace             string
+	AgentRunningOutsideManagedCluster bool
```
qiujian16: should these have json marker? also in registrationConfig, we should put it in the separate struct so we do not duplicate it, right?

zhujian7: Do you mean we put addonInstallOption inline in registrationConfig, right？

qiujian16 (commented):
/approve
/lgtm

PR #270, opened 2022-09-02
update serverURL and CABundle if ManagedCluster is created before reg…

ivan-cai:
…istration-agent running

Sometimes, ManagedCluster is created before deploy klusterlet, we want registration-agent can also update serverURL and CABundle from Klusterlet CR while it is running.

[cluster-gateway](https://github.com/oam-dev/cluster-gateway) will use  serverURL and CABundle in ManagedCluster if ClusterEndpointType is `Const` not `ClusterProxy`.

Signed-off-by: ivan-cai <caijing.cai@alibaba-inc.com>

### DIFF
--- a/pkg/spoke/managedcluster/creating_controller.go
+++ b/pkg/spoke/managedcluster/creating_controller.go
@@ -54,7 +54,7 @@ func NewManagedClusterCreatingController(
 
 func (c *managedClusterCreatingController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
 	existingCluster, err := c.hubClusterClient.ClusterV1().ManagedClusters().Get(ctx, c.clusterName, metav1.GetOptions{})
-	if err != nil && (errors.IsUnauthorized(err) || errors.IsForbidden(err) && strings.Contains(err.Error(), anonymous)) {
+	if err != nil && skipUnauthorizedError(err) == nil && strings.Contains(err.Error(), anonymous) {
 		klog.V(4).Infof("unable to get the managed cluster %q from hub: %v", c.clusterName, err)
 		return nil
 	}
@@ -64,7 +64,7 @@ func (c *managedClusterCreatingController) sync(ctx context.Context, syncCtx fac
 	}
 
 	// create ManagedCluster if not found
-	if err != nil && errors.IsNotFound(err) {
+	if errors.IsNotFound(err) {
 		managedCluster := &clusterv1.ManagedCluster{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: c.clusterName,
--- a/pkg/spoke/managedcluster/creating_controller.go
+++ b/pkg/spoke/managedcluster/creating_controller.go
@@ -53,7 +53,17 @@ func NewManagedClusterCreatingController(
 }
 
 func (c *managedClusterCreatingController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
-	_, err := c.hubClusterClient.ClusterV1().ManagedClusters().Get(ctx, c.clusterName, metav1.GetOptions{})
+	var managedClusterClientConfigs []clusterv1.ClientConfig
+	if len(c.spokeExternalServerURLs) != 0 {
+		for _, serverURL := range c.spokeExternalServerURLs {
+			managedClusterClientConfigs = append(managedClusterClientConfigs, clusterv1.ClientConfig{
+				URL:      serverURL,
+				CABundle: c.spokeCABundle,
+			})
+		}
+	}
+
+	existingCluster, err := c.hubClusterClient.ClusterV1().ManagedClusters().Get(ctx, c.clusterName, metav1.GetOptions{})
 	switch {
 	case errors.IsUnauthorized(err),
 		errors.IsForbidden(err) && strings.Contains(err.Error(), anonymous):
@@ -61,6 +71,13 @@ func (c *managedClusterCreatingController) sync(ctx context.Context, syncCtx fac
 		return nil
 	case errors.IsNotFound(err):
 	case err == nil:
+		if len(existingCluster.Spec.ManagedClusterClientConfigs) == 0 && len(managedClusterClientConfigs) > 0 {
+			clusterCopy := existingCluster.DeepCopy()
+			clusterCopy.Spec.ManagedClusterClientConfigs = managedClusterClientConfigs
+			if _, err := c.hubClusterClient.ClusterV1().ManagedClusters().Update(ctx, clusterCopy, metav1.UpdateOptions{}); err != nil {
+				return fmt.Errorf("unable to update ManagedClusterClientConfigs of managed cluster %q on hub: %w", c.clusterName, err)
+			}
+		}
 		return nil
 	case err != nil:
 		return err
@@ -71,17 +88,7 @@ func (c *managedClusterCreatingController) sync(ctx context.Context, syncCtx fac
 			Name: c.clusterName,
 		},
 	}
-
-	if len(c.spokeExternalServerURLs) != 0 {
-		managedClusterClientConfigs := []clusterv1.ClientConfig{}
-		for _, serverURL := range c.spokeExternalServerURLs {
-			managedClusterClientConfigs = append(managedClusterClientConfigs, clusterv1.ClientConfig{
-				URL:      serverURL,
-				CABundle: c.spokeCABundle,
-			})
-		}
-		managedCluster.Spec.ManagedClusterClientConfigs = managedClusterClientConfigs
-	}
+	managedCluster.Spec.ManagedClusterClientConfigs = managedClusterClientConfigs
 
 	_, err = c.hubClusterClient.ClusterV1().ManagedClusters().Create(ctx, managedCluster, metav1.CreateOptions{})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ivan-cai:
@qiujian16 plz review.

pkg/spoke/managedcluster/creating_controller.go
```diff
@@ -53,14 +54,31 @@ func NewManagedClusterCreatingController(
 }
 
 func (c *managedClusterCreatingController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
-	_, err := c.hubClusterClient.ClusterV1().ManagedClusters().Get(ctx, c.clusterName, metav1.GetOptions{})
+	var managedClusterClientConfigs []clusterv1.ClientConfig
+	if len(c.spokeExternalServerURLs) != 0 {
+		for _, serverURL := range c.spokeExternalServerURLs {
+			managedClusterClientConfigs = append(managedClusterClientConfigs, clusterv1.ClientConfig{
+				URL:      serverURL,
```
qiujian16: we need to consider the case that user update this, we should not update this back from agent

ivan-cai: > we need to consider the case that user update this, we should not update this back from agent

@qiujian16 ok, I have change the condition for updating this in https://github.com/open-cluster-management-io/registration/pull/270/files#diff-0b36934623fbbd233f2529b6c4e12a550c7a7537f8d1223070b2c2748140cdb9R74
Plz review again, thx.

pkg/spoke/managedcluster/creating_controller.go
```diff
@@ -53,14 +53,31 @@ func NewManagedClusterCreatingController(
...
 		errors.IsForbidden(err) && strings.Contains(err.Error(), anonymous):
 		klog.V(4).Infof("unable to get the managed cluster %q from hub: %v", c.clusterName, err)
 		return nil
 	case errors.IsNotFound(err):
 	case err == nil:
+		if len(existingCluster.Spec.ManagedClusterClientConfigs) == 0 && len(managedClusterClientConfigs) > 0 {
+			clusterCopy := existingCluster.DeepCopy()
+			clusterCopy.Spec.ManagedClusterClientConfigs = managedClusterClientConfigs
+			if _, err := c.hubClusterClient.ClusterV1().ManagedClusters().Update(ctx, clusterCopy, metav1.UpdateOptions{}); err != nil {
+				return fmt.Errorf("unable to update ManagedClusterClientConfigs of managed cluster %q on hub: %w", c.clusterName, err)
```
qiujian16: I do not think the agent has the permission to update this field. By default, it should not.

qiujian16: we should also ignore Unauthorized error here

pkg/spoke/managedcluster/creating_controller.go
```diff
@@ -53,14 +53,31 @@ func NewManagedClusterCreatingController(
...
+
+	existingCluster, err := c.hubClusterClient.ClusterV1().ManagedClusters().Get(ctx, c.clusterName, metav1.GetOptions{})
 	switch {
 	case errors.IsUnauthorized(err),
 		errors.IsForbidden(err) && strings.Contains(err.Error(), anonymous):
 		klog.V(4).Infof("unable to get the managed cluster %q from hub: %v", c.clusterName, err)
 		return nil
 	case errors.IsNotFound(err):
 	case err == nil:
+		if len(existingCluster.Spec.ManagedClusterClientConfigs) == 0 && len(managedClusterClientConfigs) > 0 {
```
qiujian16: do you think it is makes more sense if we check the ManagedClusterClientConfigs in clusters include the managedClusterClientConfigs, if not, merge it?

pkg/spoke/managedcluster/creating_controller.go
```diff
@@ -53,14 +53,31 @@ func NewManagedClusterCreatingController(
...
+			clusterCopy := existingCluster.DeepCopy()
+			clusterCopy.Spec.ManagedClusterClientConfigs = managedClusterClientConfigs
+			if _, err := c.hubClusterClient.ClusterV1().ManagedClusters().Update(ctx, clusterCopy, metav1.UpdateOptions{}); err != nil {
+				return fmt.Errorf("unable to update ManagedClusterClientConfigs of managed cluster %q on hub: %w", c.clusterName, err)
+			}
+		}
```
qiujian16, line 83: how about we refactor here a bit

```go
switch {
	case errors.IsUnauthorized(err):
		errors.IsForbidden(err) && strings.Contains(err.Error(), anonymous):
		klog.V(4).Infof("unable to get the managed cluster %q from hub: %v", c.clusterName, err)
		return nil
       case errors.IsNotFound(err):
                _, err = c.hubClusterClient.ClusterV1().ManagedClusters().Create(ctx, managedCluster, metav1.CreateOptions{})
               if err == nil {
                   syncCtx.Recorder().Eventf("ManagedClusterCreated", "Managed cluster %q created on hub", c.clusterName)
               }
               return err
        case err != nil:
                return err

// update client configs.
```

ivan-cai:
@qiujian16 I have refactored  the codes, and merged serverUrl. I have also ingored the Unauthorized error.
Plz review again, thx!

pkg/spoke/managedcluster/creating_controller.go
```diff
@@ -53,40 +53,80 @@ func NewManagedClusterCreatingController(
...
+	if err != nil && !errors.IsNotFound(err) {
 		return err
 	}
 
-	managedCluster := &clusterv1.ManagedCluster{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: c.clusterName,
-		},
+	// create ManagedCluster if not found
+	if err != nil && errors.IsNotFound(err) {
```
qiujian16: nit, err!=nil is not necessary

pkg/spoke/managedcluster/creating_controller.go
```diff
@@ -53,40 +53,80 @@ func NewManagedClusterCreatingController(
 }
 
 func (c *managedClusterCreatingController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
-	_, err := c.hubClusterClient.ClusterV1().ManagedClusters().Get(ctx, c.clusterName, metav1.GetOptions{})
-	switch {
-	case errors.IsUnauthorized(err),
-		errors.IsForbidden(err) && strings.Contains(err.Error(), anonymous):
+	existingCluster, err := c.hubClusterClient.ClusterV1().ManagedClusters().Get(ctx, c.clusterName, metav1.GetOptions{})
+	if err != nil && (errors.IsUnauthorized(err) || errors.IsForbidden(err) && strings.Contains(err.Error(), anonymous)) {
```
qiujian16: if errors.IsUnauthorized(err) || errors.IsForbidden(err)

do we really need to check anonymous?

Also pls add a comment here on why we return when unauthorized.

pkg/spoke/managedcluster/creating_controller.go
```diff
@@ -53,40 +53,80 @@ func NewManagedClusterCreatingController(
...
+				managedClusterClientConfigs = append(managedClusterClientConfigs, clusterv1.ClientConfig{
+					URL:      serverURL,
+					CABundle: c.spokeCABundle,
+				})
+			}
+			managedCluster.Spec.ManagedClusterClientConfigs = managedClusterClientConfigs
+		}
+
+		_, err = c.hubClusterClient.ClusterV1().ManagedClusters().Create(ctx, managedCluster, metav1.CreateOptions{})
+		if err != nil {
```
qiujian16: I think you would want to skip unauthorized here also?

It is better if we have a `func skipUnauthorizedError(err error) error`. So we can use it when run get/update/create

pkg/spoke/managedcluster/creating_controller.go
ivan-cai: > 
@qiujian16 
err != nil is necessary, because I am not check err == nil before this.

pkg/spoke/managedcluster/creating_controller.go
ivan-cai: ok

pkg/spoke/managedcluster/creating_controller.go
ivan-cai: ok

ivan-cai: @qiujian16 done!

pkg/spoke/managedcluster/creating_controller.go
qiujian16: i think 

```
if errors.IsNotFound(err) 
```
indicate err!=nil already

pkg/spoke/managedcluster/creating_controller.go
qiujian16: also use skipUnauthorizedError here

ivan-cai: > also use skipUnauthorizedError here
@qiujian16 if use skipUnauthorizedError, here will be `if err != nil && skipUnauthorizedError(err) == nil && strings.Contains(err.Error(), anonymous) ` , is it ok?

pkg/spoke/managedcluster/creating_controller.go
ivan-cai: > i think
> 
> ```
> if errors.IsNotFound(err) 
> ```
> 
> indicate err!=nil already

done

pkg/spoke/managedcluster/creating_controller.go
ivan-cai: > also use skipUnauthorizedError here

done @qiujian16

qiujian16 (commented):
/approve
/lgtm

PR #274, opened 2022-10-17
stop to discovery addon when its managed cluster is not found

skeeey:

### DIFF
--- a/pkg/hub/addon/discovery_controller.go
+++ b/pkg/hub/addon/discovery_controller.go
@@ -128,6 +128,10 @@ func (c *addOnFeatureDiscoveryController) syncAddOn(ctx context.Context, cluster
 	}
 
 	cluster, err := c.clusterLister.Get(clusterName)
+	if errors.IsNotFound(err) {
+		// no cluster, it could be deleted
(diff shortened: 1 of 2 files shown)


pkg/hub/addon/discovery_controller.go
```diff
@@ -128,8 +128,12 @@ func (c *addOnFeatureDiscoveryController) syncAddOn(ctx context.Context, cluster
 	}
 
 	cluster, err := c.clusterLister.Get(clusterName)
+	if errors.IsNotFound(err) {
+		// no cluster, it could be deleted
+		return nil
+	}
 	if err != nil {
-		return fmt.Errorf("unable to find cluster with name %q: %w", clusterName, err)
+		return fmt.Errorf("failed to list cluster with name %q: %w", clusterName, err)
```
qiujian16: why it is list here?

skeeey: roll back the msg, here should be get

ldpliu:
> /cc @ldpliu please take a look at cluster set e2e part, thanks

It looks good to me

PR #277, opened 2022-10-20
Avoid frequent CSR creation

qiujian16:

### DIFF
--- a/pkg/spoke/addon/registration_controller.go
+++ b/pkg/spoke/addon/registration_controller.go
@@ -27,9 +27,10 @@ import (
 
 const (
 	indexByAddon = "indexByAddon"
-)
 
-var AddonCSRThreshold = 10
+	// TODO(qiujian16) expose it if necessary in the future.
+	addonCSRThreshold = 10
+)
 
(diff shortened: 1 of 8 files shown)


pkg/spoke/managedcluster/registration.go
```diff
@@ -22,6 +23,11 @@ import (
 	"open-cluster-management.io/registration/pkg/hub/user"
 )
 
+const (
+	indexByCluster      = "indexByCluster"
+	clusterCSRThreshold = 10
```
skeeey, line 30: maybe a `var`, and do we need expose this arg? or add a todo here

pkg/spoke/addon/registration_controller.go
```diff
@@ -25,6 +25,12 @@ import (
 	"open-cluster-management.io/registration/pkg/helpers"
 )
 
+const (
+	indexByAddon = "indexByAddon"
+)
+
+var AddonCSRThreshold = 10
```
skeeey: do we need expose this arg? or add a todo here

skeeey:
LGTM

/cc @elgnay

PR #280, opened 2022-11-03
ensure a cluster lease is updated after the cluster is restored

skeeey:


qiujian16:
need more context on how will it resolve restore issue

skeeey:
> need more context on how will it resolve restore issue

There is a case, 
1. import a managed cluster on hub firstly
2. stop the hub, and delete the managed cluster
3. start the hub, and recreate the managed cluster (the hub can be another hub, and it have the same apiserver url and ca with the first hub)

The managed cluster available status will transition between true and unknown, this because, when we recreate  the managed cluster, the lease update routine has already stopped (we deleted the managed cluster, the lease controller capture the not found error and stop the update routine),  and the registration agent does not restart (the cluster bootstrap secret is not changed),  so the update routine will not start again

PR #282, opened 2022-11-10
fix validating webhook

ldpliu:


pkg/webhook/v1/managedcluster_validating.go
```diff
@@ -78,10 +78,12 @@ func (r *ManagedClusterWebhook) ValidateUpdate(ctx context.Context, oldObj, newO
 
 	// the HubAcceptsClient field is changed, we need to check the request user whether
 	// has been allowed to change the HubAcceptsClient field with SubjectAccessReview api
-	if managedCluster.Spec.HubAcceptsClient {
-		err := r.allowUpdateAcceptField(managedCluster.Name, req.UserInfo)
-		if err != nil {
-			return err
+	if managedCluster.Spec.HubAcceptsClient != oldManagedCluster.Spec.HubAcceptsClient {
```
qiujian16, line 81: we need to add a test for this if we do not have one.

qiujian16 (commented):
/approve
/lgtm

PR #289, opened 2022-12-05
upgrade github action

ycyaoxdu:

### DIFF
--- a/.github/workflows/go-presubmit.yml
+++ b/.github/workflows/go-presubmit.yml
@@ -22,12 +22,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 1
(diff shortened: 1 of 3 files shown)


qiujian16 (commented):
would you also help to check if we need to update kind?

ycyaoxdu:
> would you also help to check if we need to update kind?

@qiujian16 latest setup-kind is 0.5.0, latest kind is 0.17.0, updated

.github/workflows/go-presubmit.yml
```diff
@@ -117,7 +117,7 @@ jobs:
       - name: setup kind
         uses: engineerd/setup-kind@v0.5.0
```
qiujian16: do we need to upgrade this version?

ycyaoxdu: @qiujian16  v0.5.0 is latest https://github.com/engineerd/setup-kind/tags

qiujian16: e2e fails..

PR #291, opened 2022-12-15
roll back code to support lower version k8s

skeeey:


skeeey:
rollback the pr https://github.com/open-cluster-management-io/registration/pull/275/files

PR #295, opened 2023-02-09
quick restart container when kubeconfig changed

zhujian7:

### DIFF
--- a/pkg/spoke/spokeagent.go
+++ b/pkg/spoke/spokeagent.go
@@ -2,32 +2,20 @@ package spoke
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


pkg/spoke/spokeagent.go
```diff
@@ -406,6 +422,81 @@ func (o *SpokeAgentOptions) RunSpokeAgent(ctx context.Context, controllerContext
...
+	s.cancel()
+}
+
+func (s *probeServer) WaitForStop() {
+	<-s.closed
+	klog.Info("probe server was stopped")
+}
+
+// serveHealthProbes serves health probes and configchecker.
+func serveHealthProbes(ctx context.Context, healthProbeBindAddress string,
```
qiujian16, line 465: we already have healthz in registration, why we need to start a new one?

PR #301, opened 2023-02-23
auto approve bootstrap csr

skeeey:


pkg/hub/csr/reconciler.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (b *csrBootstrapReconciler) Reconcile(ctx context.Context, csr csrInfo, approveCSR approveCSRFunc) (reconcileState, error) {
+	// Check whether current csr is a valid spoker cluster csr.
+	if valid, _ := validateCSR(csr); !valid {
+		klog.V(4).Infof("CSR %q was not recognized", csr.name)
+		return reconcileStop, nil
+	}
+
+	// Check whether current csr can be approved.
+	if !b.approvalUsers.Has(csr.username) {
```
qiujian16, line 126: do we also need to set accepthubclient=true?

skeeey: yeah, we may need it, so we will have a whole auto approval process for a spoke

skeeey: accept a cluster if the cluster is not accepted

qiujian16:
/approve

Anyone else you would need to review this PR. I think we should also refactor this controller to leverage generics as a followup

pkg/hub/manager.go
```diff
@@ -37,8 +38,26 @@ import (
...
+
+// NewHubManagerOptions returns a HubManagerOptions
+func NewHubManagerOptions() *HubManagerOptions {
+	return &HubManagerOptions{}
+}
+
+// AddFlags registers flags for manager
+func (m *HubManagerOptions) AddFlags(fs *pflag.FlagSet) {
+	features.DefaultHubMutableFeatureGate.AddFlag(fs)
+	fs.StringArrayVar(&m.ClusterAutoApprovalUsers, "--cluster-auto-approval-users", m.ClusterAutoApprovalUsers,
```
morvencao, line 54: should remove `--` here, otherwise the flag will be:
```bash
Flags:
      ----cluster-auto-approval-users stringArray   A list of reachable spoke cluster api server URLs for hub cluster.
```

PR #303, opened 2023-03-09
Remove: work directory in github action.

xuezhaojun:


.github/workflows/go-presubmit.yml
```diff
@@ -23,25 +19,19 @@ jobs:
...
+
   verify-deps:
     name: verify-deps
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
         uses: actions/checkout@v3
-        with:
-          fetch-depth: 1
-          path: go/src/open-cluster-management.io/registration
```
skeeey, line 44: we may keep this to avoid clone the whole repo?

.github/workflows/go-presubmit.yml
```diff
@@ -23,25 +19,19 @@ jobs:
     steps:
       - name: checkout code
         uses: actions/checkout@v3
-        with:
-          fetch-depth: 1
```
xuezhaojun, line 27: The default `fetch-depth` is 0. https://github.com/actions/checkout/blob/ac593985615ec2ede58e132d2e21d2b1cbd6127c/README.md?plain=1#L78

.github/workflows/go-presubmit.yml
xuezhaojun: The default fetch-depth is also 1. https://github.com/actions/checkout/blob/ac593985615ec2ede58e132d2e21d2b1cbd6127c/README.md?plain=1#L78

PR #310, opened 2023-04-04
Fix: deny accept request when namespace is terminating.

xuezhaojun:
Related to: https://github.com/open-cluster-management-io/registration-operator/pull/342

### DIFF
--- a/pkg/webhook/v1/managedcluster_validating.go
+++ b/pkg/webhook/v1/managedcluster_validating.go
@@ -19,6 +19,8 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	corev1 "k8s.io/api/core/v1"
 )
 
 var _ webhook.CustomValidator = &ManagedClusterWebhook{}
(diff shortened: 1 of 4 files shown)


pkg/webhook/v1/managedcluster_validating.go
```diff
@@ -173,6 +184,29 @@ func (r *ManagedClusterWebhook) allowUpdateAcceptField(clusterName string, userI
 	return nil
 }
 
+// validateClusterNamespace checks the cluster namespace, if the namespace is terminating, reject the accept request.
+func (r *ManagedClusterWebhook) validateAcceptByClusterNamespace(clusterName string) error {
+	clusterNamespace, err := r.kubeClient.CoreV1().Namespaces().Get(context.TODO(), clusterName, metav1.GetOptions{})
+	if err != nil {
```
qiujian16, line 193:
```go
if apierrors.IsNotFound(err) {
	return nil
}
if err != nil {
  return ...
}
```

xuezhaojun: Fixed.

xuezhaojun: Please take another look when you have time! Thanks!

Issue #311, opened 2023-04-18
Specify ExpirationSeconds in CSR request


youhangwang:
v1CSRControl create a csr request without ExpirationSeconds. the expiration time of the signed certificate depends on the --cluster-signing-duration in controller manager.

can we add a expirationSeconds param in the function to specify a expiration time to create a csr?

https://github.com/open-cluster-management-io/registration/blob/fc792f3816ed1e2f2d0609a5c0ac69da32ed2ab9/pkg/clientcert/certificate.go#L216

skeeey:
yeah, we may add an option for registration agent to configure this

/cc @qiujian16

qiujian16:
+1

youhangwang:
I can make this change as the first contribution to OCM proj ;)

PR #312, opened 2023-04-18
add expirationSeconds in CSROption to create CSR

youhangwang:


youhangwang:
https://github.com/open-cluster-management-io/registration/issues/311

skeeey:
@youhangwang thanks for your contributions!

Would you also help to add an option in https://github.com/open-cluster-management-io/registration/blob/main/pkg/spoke/spokeagent.go#L56

and expose the option with flags https://github.com/open-cluster-management-io/registration/blob/main/pkg/spoke/spokeagent.go#L410

youhangwang:
yes, sure. Thanks for the review @skeeey 

please help to review the flag name.

pkg/spoke/spokeagent.go
```diff
@@ -425,6 +428,8 @@ func (o *SpokeAgentOptions) AddFlags(fs *pflag.FlagSet) {
 		"The period to check managed cluster kube-apiserver health")
 	fs.IntVar(&o.MaxCustomClusterClaims, "max-custom-cluster-claims", o.MaxCustomClusterClaims,
 		"The max number of custom cluster claims to expose.")
+	fs.Int32Var(&o.ClientCertExpirationSeconds, "client-cert-expiration-seconds", o.ClientCertExpirationSeconds,
```
skeeey, line 431: I think we need validate this value in the `Validate` function https://github.com/open-cluster-management-io/registration/blob/main/pkg/spoke/spokeagent.go#L431, if this value is specified, we need require a min expiration 600s

youhangwang: added the validation for client-cert-expiration-seconds

skeeey:
LGTM

@youhangwang we need a sign-off for your commits

/cc @qiujian16
please also take a look

qiujian16:
/lgtm

Thanks for your contribution. @skeeey we need to also think how to expose these flags from operator API, same as --approve-user-list flag.

youhangwang:
@skeeey @qiujian16 

The minimum valid value for expirationSeconds in CSR spec is 600.

But for the clientCertificateController in registration, 600 is too short:
1. clientCertificateController to check if cert is expired every 5min.
2. kubeclient resync the client cert every 5min.
3. clientCertificateController will refresh the cert if the cert has less than a random percentage range from 20% to 25% of its life remaining.

so the minimum expiration second would be (5 * 60)/(1/5) = 1500s for clientCertificateController. otherwise, the kube client will use a expired client cert to send request to the remote cluster.


Considering to add some margin. 1h(3600s) could be a good choice. Any thoughts on this?

skeeey:
@youhangwang you are right, I forget the cert renew time,  I think we need increase the min expired time

youhangwang:
https://github.com/open-cluster-management-io/registration/pull/320

